PR #141, opened 2017-08-08
remove zs (resolves #136)

### LINKED ISSUE #136: remove "z" from channel.zmag, zmax, zmin, znull
the "z" in these attributes is some kind of weird holdover from the days of fscolors, and I don't think we should keep it moving forward---I want to keep things more consistent throughout the package

the old attributes will be kept for now, but they will raise a depreciation warning if used

we will remove them in the next major release

untzag:


untzag:
Also resolves #142.

Issue #137, opened 2017-08-06
illegal characters in axis names


darienmorrow:
WrightTools allows me to create an axis with name `'w2=w1'` but this is not a python recognizable object attribute due to the character `=`. 

This is annoying because the way one generally gets the points of an axis is to write `d.w2.points` but in this case, `d.w2=w1.points` is not allowed by python. So one must resort to `d.axes[0].points`.

darienmorrow:
[000.zip](https://github.com/wright-group/WrightTools/files/1212410/000.zip)
Here is a .zip file that contains a .data file which has an example of an illegal axis name.

PR #147, opened 2017-08-10
documentation

untzag:


ksunden:
That regex should be `[a-zA-Z_][a-zA-Z0-9_]*`

untzag:
you're exactly correct @ksunden, and in fact I just added a method in kit for that to solve #137

PR #148, opened 2017-08-10
pep8speaks configuration (closes #145)

### LINKED ISSUE #140: Tuning sometimes fails to auto create fit, probably because of cutoff factor problems
When the tunetest curve generator tries to make a fit, it sometimes fails. Jonathan guesses it has to do with the cutoff factor. We should consider having a smart way to handle that error and either try other values for the cutoff parameter or do something else user friendly.

### LINKED ISSUE #145: .pep8speaks.yml
https://github.com/OrkoHunter/pep8speaks#configuration

@ksunden could you add a yml with the appropriate configuration options?

Thanks :kissing_heart:

ksunden:
This is very basic for now, I did not mess with the text that appears, but that can be done later. Mostly I wanted to get the max line length updated (and I like count)


ksunden:
I believe that the failed check is because it was only checking the changed files, which have no tests as its a configuration file. Does this make sense with the Travis settings @untzag?

Issue #138, opened 2017-08-07
wt.data.join all nan slices


jhandali:
Using the function wt.data.join creates a data object combining multiple files, but it creates all nan slices at the boundary between the parent objects.

PR #152, opened 2017-08-10
join

untzag:
please pay special attention to the docstring of data.join---I am not sure what the correct strategy is for documenting kwargs so I did my best, but I'm open for suggested improvements

note that I added the bug from #138 as a test, so it shouldn't be able to sneak back in unnoticed

feel free to delete branch after merging :kissing_heart:


ksunden:
Looks good, hopefully the pep8speaks bot is only so noisy because the config file isn't in your branch (I think that's why, any way)

Issue #153, opened 2017-08-10
data `from_text` 


ksunden:
Currently this seems to be a rather specific text format reader for Brunold rRaman files

I feel like this could be made more general to any tab separated file, but think that it needs to be considered exactly what specifications should be there.

If we expand, it should not check for expected filetypes, as any plaintext file will do.

There is also a portion in the verbose output that references `datas` a variable that does not exist

untzag:
I have been thinking of getting rid of `from_text`. I much rather have each from method be entirely separate. I know that this will mean more lines of code, but I think that portability is more important here.

Anyway, there are great tools like `np.genfromtext` that do most of what we want anyway.

ksunden:
I can agree to that, regardless, `from_text` which expects specifically rRaman files is confusing

ksunden:
@neffmallon has mentioned that from_text is used for files from the raman instrument downstairs

This should likely have its own `from` method, but should be done concurrently with removing `from_text`

neffmallon:
*from the IR instrument downstairs.

ksunden:
Yeah, sorry, I mistyped that, read it back a little bit ago, and thought that sounded incorrect

untzag:
which IR instrument? we have `from_Tensor27`

ksunden:
I actually think maybe we want to keep `from_text` in some form, but maybe rename it to something like `from_ascii_xy` and put some specificity into what we expect

Maybe with support for headers which specify channel/axis names, though I think it would be limited to single axis, potentially with multiple channels though

There is a helpful list provided by Essential FTIR of many common spectroscopy formats:
http://www.essentialftir.com/filetypes.html

Here, they do list ascii (x,y) pairs, which seems to be a generic enough format, which we could pretty easily support, but need to be clear that is what it means

ksunden:
Keeping it around with some flexible options (perhaps kwargs for channel names, etc) would also make it a powerful way to create many of the common formats with minimal effort

untzag:
I guess that sounds reasonable, but it does worry me. I really don't trust users to remember the details of specific file types, which I why I much prefer dedicated from methods whenever possible. I worry that everyone would just start using `from_ascii`, never remembering to use the dedicated methods.

I guess I'm OK with such a method existing, as long as it doesn't get advertised for general use.

untzag:
`from_text` has been removed (see #233)

PR #163, opened 2017-08-11
Adds a rRaman example, renames and changes data.from_rRaman to data.from_BrunoldrRaman

### LINKED ISSUE #162: Example of Brunold Group rRaman data loading and plotting

neffmallon:
resolves #162, lessens dependence on from_text (see #153)


WrightTools/data.py
```diff
@@ -1934,6 +1934,62 @@ def from_text(filepath, name=None, verbose=True):
...
+        else:
+            name = filepath.split('//')[-1].split('.')[0]
+            data = Data([axis], [signal], source='txt file', name=name)
+    # finish
+    if verbose:
+        print('{0} data objects successfully created from file:'.format(len(indicies)))
+    return data
+
+
+def from_BrunoldrRaman(filepath, name=None, verbose=True, trash_pixels=56):
```
ksunden: If trash_pixels is a kwarg, it should be listed in the docstring

WrightTools/data.py
```diff
@@ -1934,6 +1934,62 @@ def from_text(filepath, name=None, verbose=True):
...
+    if isinstance(filepath, type([])) or isinstance(filepath, type(np.array([]))):
+        return [from_BrunoldrRaman(f) for f in filepath]
+    if not os.path.isfile(filepath):
+        raise wt_exceptions.FileNotFound(path=filepath)
+    if not filepath.endswith('txt') and not filepath.endswith('dpt'):
+        wt_exceptions.WrongFileTypeWarning.warn(filepath, 'txt or dpt')
+    # import array
+    lines = []
+    with open(filepath, 'r') as f:
+        for i in range(trash_pixels):
```
ksunden: @untzag What is your opinion on doing this kind of data removal in a `from` method?

The particular instrument that the Brunold group has apparently produces junk data at the beginning of the file.  It is not clear what the cause of these junk values is, but they are not valuable to the data object itself

WrightTools/data.py
```diff
@@ -1934,6 +1934,62 @@ def from_text(filepath, name=None, verbose=True):
...
+            line = f.readline()
+        while True:
+            line = f.readline()
+            if line == '\n' or line == '':
+                break
+            else:
+                clean = line[:-2]  # lines end with ',/n'
```
ksunden: The datafile provided does not have commas, are we losing the precision here 

(Admittedly in the example file, the truncated digit would always be zero, but I do not know if the precision shown is always the case)

WrightTools/data.py
neffmallon: I went upstairs and talked to the Brunold group and it looks like the negative pixel problem has been either fixed or never existed. I've taken the trash_pixels out of this code and changed the example code too.

untzag:
thanks so much @neffmallon

neffmallon:
old branch is still around, should it be deleted meow?

untzag:
I don't see it @neffmallon

Issue #151, opened 2017-08-10
from_shimadzu python3 compliance


ksunden:
the `from_shimadzu` import function uses `raw_input` which was renamed as `input` in python3. This will break if trying to use the method  with a file which does not end in 'txt'.

Workaround: 
Rename files so that the condition for calling raw_input is not met: i.e. add .txt as the file extension

ksunden:
Perhaps the solution is to trust the user and remove the file path checking all together

Issue #155, opened 2017-08-10
Links in docstrings


ksunden:
I prefer the rST link format where the links are separated from the hyperlink text

This makes lines shorter (silencing flake8 warnings often, unless urls are particularly long)
And also makes it (in my opinion) easier to read, as the urls clog up the view when most of the time I am not interested in them unless I am explicitly looking for the link.

For short urls, it is still appropriate to have it in line, as those are not too distracting, and can be useful

The link url is available, and the syntax is such that it is clear a link is indicated, and the URL should be relatively close by.

untzag:
I agree, and move that we bring all links to the bottom of the docstring.

@ksunden can you just go through and do this all at once?

ksunden:
Note the rST format  requires the links to be in the same block, so I am putting them closer to the actual reference to that link

ksunden:
I am mildly confused about the rST linking format. You can refer to the same thing in multiple docstrings, but can only have the actual link url in one place (if you want the same text to show up, e.g. griddata in data.py)

untzag:
@ksunden is it possible to have one 'repository' of urls at the top or bottom of each module (for clarity and cleanliness)?

ksunden:
perhaps, I'll look into it, perhaps in the docstring at the file level

ksunden:
Putting in the file level docstring works

does this make sense as a way to proceed to you @untzag

ksunden:
See PR #166 which has been updated so that artists.py does this

If we are happy with artists, I will do the same for the remaining files

PR #164, opened 2017-08-11
Flake8 

### LINKED ISSUE #159: Reduce flake8 warnings
We currently have over 100 warnings (down from thousands prior to autopep8 usage, but still)

This will take several iterations, and may not be able to get it down to zero for a few releases

See also:

#158  -  current causes several warnings
#155  -  current causes lines to be too long, some links may not be short enough in general, but closing #155 would help several places where lines are too long

ksunden:
Many fixes to reduce flake8 warnings significantly

Not perfect, still around 30 or so, many of which are lines that are too long while being annoying or ugly to split.

Closes #159, new issues should be spawned for any remaining problems

Next step is #155 which will help several of the line too long complaints


ksunden:
Also closes #151

untzag:
thanks so much @ksunden

PR #166, opened 2017-08-14
Links in Docstrings

ksunden:


WrightTools/kit.py
```diff
@@ -637,7 +644,7 @@ def read_headers(filepath):
 
 
 def write_h5(filepath, dictionary):
-    """ Save a python dictionary into an `HDF5 <https://www.hdfgroup.org/HDF5/doc/H5.intro.html>`_
+    """ Save a python dictionary into an `HDF5`_
```
untzag: does this link inherit from the docstring earlier in the file?

ksunden: It does, and it throws errors if you try to put the same tag in but redefine it...

I don't like not having the link in the docstring itself, but I'm not sure how to resolve that at this time

ksunden:
I found out rST does allow hyperlinks to be split on multiple lines (and ignores whitespace) This will remove the majority of remaining line length errors

WrightTools/kit.py
ksunden: One thing we could do is use anonymous hyperlink in this case, i.e. `__` and then have 
```
.. __: hdf5-write_
.. _hdf5-write: https://hdf5.org/path/to/index.html
```

ksunden: This would allow us to have unique names for the links themselves in underlying rST

untzag: in this case you're the expert @ksunden---whatever you think is the best strategy

untzag:
@ksunden are you working in this branch or shall I merge now?

ksunden:
I plan on cleaning up a few more bits, please leave this open for the time being

PR #168, opened 2017-08-14
Doilinks

ksunden:


ksunden:
Looking into using `bibtex` and `sphinxcontrib-bibtex` to get full citations in a standardized form. I do want to make sure that it is still a simple click away, so that may need some tinkering, depending on the format that is used by default. 

I think that should be opened as a separate issue, and this to be merged with its current functionality as is for the first step

PR #186, opened 2017-08-21
Change 'Wright Group' to 'WrightTools'

ksunden:


untzag:
great catch---thanks @ksunden

PR #188, opened 2017-08-24
simplify distribution

untzag:


untzag:
@ksunden please pay special attention to upload.sh---it is my first real bash script and I probably made some obvious errors

scripts/upload.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
+cd ..
```
ksunden: This is dangerous -- scripts should not be required to be run in a certain directory

scripts/upload.sh
```diff
@@ -0,0 +1,30 @@
...
+    exit 1
+fi
+# build
+printf "\033[1;35mbuilding...\033[0m\n"
+python setup.py sdist
+python setup.py bdist_wheel --universal
+# upload
+printf "\033[1;35muploading...\033[0m\n"
+while true; do
+        read -p "upload to PyPI? [y/n]: " yn
```
ksunden: I typically like these scripts to have a default (indicated with a capital letter in the prompt) so that the most common action is just an enter press away.

Now, if you want to make sure that uploads are intentional things, no default is okay

untzag: I think in this case I like having a required answer here, because uploading is really something I want people to be conscious of.

If we went with a default I would make the default N.

Issue #183, opened 2017-08-18
kit.TimeStamp.hms display on web page


ksunden:
current rST source is 
```
``HH:MM:SS``
```

For an unknown reason this get transliterated to `HH -- MM -- SS` when it is displayed in the HTML file

I have been unable to track down why, it is not documented behavior in the rST standard

The same was not true when I tried on the online [editor](http://rst.ninjs.org/?theme=basic#YGBISDpNTTpTU2Bg)

Switching to single backticks in the source changed the color, and changed it to a single dash

Using two colons in a row, they showed up as expected

ksunden:
Annoyingly, It seems to work exactly as expected if the HH:MM:SS is not on the first line...

ksunden:
Works fine with numpydoc as the parser - bug in napoleon

ksunden:
I created a branch, hmsColon, which is here specifically to preserve the issue as documented.

Once sphinx-doc/sphinx#4016 is closed, that branch should be deleted

ksunden:
The upstream bug was closed... I have no real interest in reverting this docstring at this time.

Will remove the old branch

PR #191, opened 2017-08-25
Put the HH:MM:SS on a new line (see #183)

ksunden:


ksunden:
Issue #183 should be open until the corresponding sphinx issue is closed

PR #214, opened 2017-08-31
Issue199

kjczech:
I have added two items for which I am creating a pull request:

1. added WrightTools.data.from_spc130
2. added the directory ..\WrightTools\WrightTools\datasets\spc130 and with an example data file in it to check formatting, should a user have trouble importing a spc130 data file.


WrightTools/data.py
```diff
@@ -2861,9 +2861,79 @@ def from_shimadzu(filepath, name=None, verbose=True):
...
+    """Create a data object from a SPC-130 TCSPC an exported
+    comma-delimited (.asc) file within the SPCM 9.75 software.
+
+    Parameters
+    ----------
+    filepath : string
+        Path to SPC-130 output file (.asc).
+    name : string (optional)
+        Name to give to the created data object. If None, filename is used.
+        Default is None.
```
untzag: please document delimiter kwarg

WrightTools/data.py
```diff
@@ -2861,9 +2861,79 @@ def from_shimadzu(filepath, name=None, verbose=True):
...
+                  "in wt.data.from_spc130() call." % delim_dict[delimiter])
+        for delimiter in delim_args:
+            arr = np.genfromtxt(filepath,
+                                skip_header=10, skip_footer=1,
+                                delimiter=delimiter).T
+            if not np.any(np.isnan(arr)):
+                if verbose:
+                    print("Error resolved: file is %s-delimited."
+                          % delim_dict[delimiter])
+                break
```
untzag: personally I would use warnings as opposed to print conditional to `verbose` in this if block

WrightTools/data.py
```diff
@@ -2861,9 +2861,79 @@ def from_shimadzu(filepath, name=None, verbose=True):
...
+    if np.any(np.isnan(arr)):
+        print("Error unresolved: Please check that your file "
+              "is formatted properly. An example of SPC-130 file format "
+              "can be found in ..\WrightTools\WrightTools\datasets\spc130"
+              "\n")
+        print('data object not created!\n')
```
untzag: you should raise an exception here---the user should not be allowed to proceed

`raise RuntimeError('your message here')

WrightTools/data.py
untzag: also if the default is `','` it's probably best to put that into the kwarg (instead of `None`)

untzag:
@kjczech in addition to the requested changes could you fully integrate your dataset?
1) add an import test
2) add dataset properties to documentation

thanks so much for the pr :kissing_heart:

WrightTools/data.py
```diff
@@ -2861,9 +2861,79 @@ def from_shimadzu(filepath, name=None, verbose=True):
     return data
 
 
+def from_1130(filepath, name=None, delimiter=None, verbose=True):
+    """Create a data object from a SPC-130 TCSPC an exported
+    comma-delimited (.asc) file within the SPCM 9.75 software.
```
ksunden: Docstring summary should fit on one line, additional detail may be included between the summary and Parameters section (with blank lines separating the parts)

See [Numpy](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#docstring-standard)

Also, [pydocstyle](http://www.pydocstyle.org) will help to ensure that the style is kept

WrightTools/data.py
ksunden: I agree, here, warnings are here for just such an occasion (though I may put the resolved in a logging layer like INFO, if we had such logging implemented)

WrightTools/data.py
ksunden: Yes, then remove the else block, there should be at least one return that is not in a conditional block, usually (some old-school programmers will say that there should be exactly one return statement for a given method, but I think that harms more often than it helps readability)

WrightTools/data.py
```diff
@@ -2861,9 +2861,79 @@ def from_shimadzu(filepath, name=None, verbose=True):
...
+        if verbose:
+            print('data object created!\n')
+        # return
+        return data
+
+
 def from_Tensor27(filepath, name=None, verbose=True):
     """Create a data object from a Tensor27 FTIR file.
 
+
```
ksunden: Please remove this additional blank line

ksunden:
As @untzag stated, please add your data format to the documentation, specifically the table in 
https://github.com/wright-group/WrightTools/blob/development/docs/datasets.rst#datasets

untzag:
@kjczech we're committing tomorrow so you've got to finish this today or wait until 2.14.0

either option is fine :smile:

WrightTools/data.py
```diff
@@ -2900,63 +2900,63 @@ def from_spc130(filepath, name=None, delimiter=',', verbose=True):
...
-    if verbose:
-        print('data object created!\n')
-    # return
-    return data
+        if np.any(np.isnan(arr)):
+            try:
+                error = ('\nUnable to load data file.\nData object not created!'
+                    'Please check that your file is formatted properly.\n\n'
+                    'An example of the SPC-130 file format can be found in'
+                    '<..\WrightTools\WrightTools\datasets\spc130>')
```
ksunden: You can refer to this as `WrightTools.datasets.spc130`, these datasets are distributed, and that attribute should point to the file path to the data file

WrightTools/data.py
```diff
@@ -2900,63 +2900,63 @@ def from_spc130(filepath, name=None, delimiter=',', verbose=True):
...
-    data = Data([x_axis], [signal], source='SPC_130',
-                name=name)
-    data.attrs.update(headers)
-    
-    if verbose:
-        print('data object created!\n')
-    # return
-    return data
+        if np.any(np.isnan(arr)):
+            try:
```
ksunden: You should not use try/except to always raise, instead use `raise ValueError(error)`

error should be a single string, can be on multiple lines, the first line will be highlighted more than others.

WrightTools/data.py
```diff
@@ -2900,63 +2900,63 @@ def from_spc130(filepath, name=None, delimiter=',', verbose=True):
...
-        "Trying other delimiters in wt.data.from_spc130() call."\
-        %delim_dict[delimiter])
-        
-        warnings.warn(message)
-        
+
+        warning_message = ('\n\nfile is not %s-delimited!\n'\
+        'Trying other delimiters in wt.data.from_spc130() call.'\
+                                                       %delim_dict[delimiter])
+        print(warning_message)
```
ksunden: This should be done with `warnings.warn`

WrightTools/datasets/__init__.py
```diff
@@ -53,29 +53,32 @@ def from_directory(self, dirname, prefix=''):
...
+KENT._from_directory(os.path.join(here, 'KENT', 'PbSe 2D delay A'))
+KENT._from_directory(os.path.join(here, 'KENT', 'PbSe 2D delay B'))
 
 PyCMDS = DatasetContainer()
-PyCMDS.from_files('PyCMDS')
+PyCMDS._from_files('PyCMDS')
 
-Cary50 = DatasetContainer()
-Cary50.from_files('Cary50')
+spcm = DatasetContainer()
```
ksunden: You got rid of Cary50...

PR #219, opened 2017-09-01
Issue158

ksunden:


ksunden:
Note, I have only tested this on linux, not windows or mac. If we would like to test on other platforms, let me know, and I will organize/guide those tests
Also note, as of appdirs 1.3.0, there exists a user_config_dir method, which points to `~/.config` on linux, where user_data_dir points to `~/.local/share`. On mac and windows, the two methods return the same path, and I went with the longer standing method for compatibility with older versions (OK, I did that because I was looking at older documentation, and decided not to fix for compatibility, but see below for discussion on restricting version)

Older versions of appdirs points data_dir to `~/.config`

One more note: the [changelog](https://pypi.python.org/pypi/appdirs/1.4.3) states that python 3.6 support came in version 1.4.2 (with bug fix in 1.4.3). Should we require this version for that reason? @untzag 

If we do require that version, I would want to think whether this belongs in config or .local/share on linux, and use the appropriate method.

untzag:
yeah I would like to ensure python 3.6 support actually

@ksunden can you update?

PR #225, opened 2017-09-03
similar packages (resolves #213)

### LINKED ISSUE #213: similar packages
would like to have a similar-packages page on wright.tools

[gridded](https://noaa-orr-erd.github.io/gridded/index.html)
[xarray](http://xarray.pydata.org/en/stable/)

untzag:


untzag:
I'm not sure why the checks failed initially---they are passing now

PR #226, opened 2017-09-04
max, min, null, and mag methods (resolves #193)

### LINKED ISSUE #193: max and min methods

untzag:
I think I got all of them, but I could be wrong


ksunden:
Also, fit.py contains 
```
739-        for i in range(len(self.function.params)):
740-            # give the data all at once
741-            channel = self.outs.channels[i]
742-            values = channel.values
743:            channel.null = 0
744-        self.outs._update()
745-        return self.outs
746-
747-
```

Issue #218, opened 2017-09-01
Spectral delay correction process wigner ignores large negative values (in the thousands)


wright-stuff:
There are clip statements which clip the width channel and the mean channel to (0,500) and (-1000, 1000), respectively. I had to comment both of these out to process a wigner.

https://drive.google.com/open?id=0BzJTClorMBuwMjdnaTFwN09lcWM

Hotfixed on the picosecond system, just commented out, we should discuss this further.

darienmorrow:
This issue is related to [#184](https://github.com/wright-group/WrightTools/issues/184).

ksunden:
https://github.com/wright-group/WrightTools/blob/7d44922ee8401b32f0e83f61c4bed28f0d4aefa7/WrightTools/tuning/spectral_delay_correction.py#L58-L70

PR #229, opened 2017-09-05
process_wigner fixes

untzag:
process_wigner now has two cutoff factors exposed to user:
1) global_cutoff_factor
2) slice_cutoff_factor

the SDC scans that failed to process for @darienmorrow are now in the testing suite

turns out that the real problem was that there were channels without ANY data in those scans
this works poorly with share_nans
I have added another parameter to fitter.fit to address this
@ksunden @darienmorrow let's chat about this sometime


WrightTools/tuning/spectral_delay_correction.py
```diff
@@ -51,15 +70,21 @@ def process_wigner(data_filepath, channel, control_name,
...
-    outs = fitter.run(channel_index)
+    outs = fitter.run(channel_index, propagate_other_channels=False, verbose=False)
     # clean
     # remove the edges because they are badly behaved...
     # should probably do something more sophisticated...
-    outs.amplitude.clip(outs.amplitude.max() * 0.1, outs.amplitude.max())
-    #outs.amplitude.clip(outs.amplitude.max()*0.04, outs.amplitude.max())
     outs.amplitude.values[0] = np.nan
     outs.amplitude.values[-1] = np.nan
     outs.width.clip(0, 500)
```
ksunden: This line and the one that clips based on the mean cause problems on the picosecond system (See #218). Now would probably be a good time to evaluate their utility. The ability to specify units as a parameter may help, but I'm not sure, would have to test.

ksunden: That is referring to `outs.width.clip(0,500)`

ksunden:
I don't have anything immediately, but do want to test it out with a dataset we had a failure on last week, and so will refrain from approving the changes until tomorrow. I do agree that @untzag @darienmorrow and I should get together and chat about this method some. Now may be a good time to review any pain points associated with it.

untzag:
I guess that loosening those restraints might be in order

@ksunden you might try changing delay_units to `'ps'`?

untzag:
resolves #218

WrightTools/tuning/spectral_delay_correction.py
```diff
@@ -87,8 +87,9 @@ def process_wigner(data, channel, control_name, offset_name, coset_name, color_u
     # should probably do something more sophisticated...
     outs.amplitude.values[0] = np.nan
     outs.amplitude.values[-1] = np.nan
-    outs.width.clip(0, 500)
-    outs.mean.clip(-1000, 1000)
+    width = data.channels[0].max() - data.channels[0].min()
+    outs.width.clip(0, width)
+    outs.mean.clip(data.channels[0].min(), data.channels[0].max())
```
ksunden: These should be axes[0], not channels[0]

PR #241, opened 2017-09-07
break up data module (resolves #240)

### LINKED ISSUE #240: break up data module

untzag:


ksunden:
I expected all the Axis, Data, Collection, etc to be in their own file

Doing this would make us have more imports in the other files (potentially), but would be more explicit in imports.

I'm not a huge fan of module `*` imports, I would prefer keeping the flake8 check, and explicitly importing that which we need.

untzag:
@darienmorrow thank you for your insightful comment

untzag:
@ksunden yes, getting rid of * is a good idea

I don't really see the point of breaking up main.py---`Axis`, `Channel`, `Collection`, `Data`, and `join` are all so tightly coupled

I can see breaking out `Collection` and `join` I guess

let's talk about this

WrightTools/data/colors.py
```diff
@@ -0,0 +1,340 @@
...
+# --- define --------------------------------------------------------------------------------------
+
+
+__all__ = ['from_COLORS']
+
+
+# --- helpers -------------------------------------------------------------------------------------
+
+
+def discover_dimensions(arr, dimension_cols, verbose=True):
```
ksunden: Perhaps these should be _discover_dimensions, otherwise a) this will be a user facing method that is really intended for internal use only, and b) There will be namespace issues with having two copies of this which are both imported in __init__

untzag: note that `discover_dimensions` is not in `__all__`, so no need to worry about namespace issues as far as I know

PR #269, opened 2017-09-14
symmetric sqrt (resolves #267)

### LINKED ISSUE #267: symmetric sqrt
tired of writing and rewriting symmetric sqrt operations in everything that I do

untzag:


WrightTools/kit.py
```diff
@@ -1056,6 +1056,40 @@ def __init__(self, xi, yi, k=3, s=1000, ignore_nans=True):
...
+        Symmetric square root of arr.
+    """
+    if hasattr(x, '__len__'):
+        x = np.array(x)
+        number = False
+    else:
+        x = np.array([x])
+        number = True
+    factor = np.ones(x.shape)
+    factor[x < 0] = -1
```
ksunden: please use [np.sign](https://docs.scipy.org/doc/numpy/reference/generated/numpy.sign.html)

This will also allow you to transparently pass numbers, without any if/else blocks

ksunden: I just realized your test is exactly that... Why all the overhead in the function?

PR #274, opened 2017-09-15
refactor flatten_list (resolves #272)

### LINKED ISSUE #272: flatten list

untzag:


WrightTools/kit.py
```diff
@@ -1153,36 +1152,43 @@ def array2string(array, sep='\t'):
     return string
 
 
-def flatten_list(l):
-    """Flatten an irregular list.
+def flatten_list(items, seqtypes=(list, tuple), in_place=True):
```
ksunden: Do we wish for the default to be in place? This is not consistent with previous behavior.

WrightTools/kit.py
```diff
@@ -1153,36 +1152,43 @@ def array2string(array, sep='\t'):
...
-                Temp.extend(element)
-                keepChecking = True
-            else:
-                Temp.append(element)
-        listIsNested = keepChecking  # determine if outer loop exits
-        lis = Temp[:]
-    return lis
+    if not in_place:
+        items = items[:]
+    for i, x in enumerate(items):
```
ksunden: please use `_` for the unused variable `x`

WrightTools/kit.py
untzag: yeah, I thought about that and I think I do want to move to in place

PR #275, opened 2017-09-15
call plot_colorbar in mpl_2D (resolves #263)

### LINKED ISSUE #263: mpl_2D colour bar is pointy at the ends when pixelated = Off
The upper and lower extremities of the colour bar are pointed rather than flat when pixelated is turned off. The colour bar resembles a sharpened stake used to ward off enemy cavalry charges in the Middle Ages.

untzag:
@ksunden please pull and verfy that the issue is fixed on your machine

### DIFF
--- a/WrightTools/artists.py
+++ b/WrightTools/artists.py
@@ -904,7 +904,7 @@ def make_colormap(seq, name='CustomMap', plot=False):
             cdict['blue'].append([item, b1, b2])
     cmap = mplcolors.LinearSegmentedColormap(name, cdict)
     if plot:
-        plot_colormap(cmap)
+        plot_colormap_components(cmap)
     return cmap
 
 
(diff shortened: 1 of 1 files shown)


WrightTools/artists.py
```diff
@@ -2195,16 +2195,9 @@ def plot(self, channel=0,
             for xi, yi, kwargs in self._onplotdata:
                 subplot_main.plot(xi, yi, **kwargs)
             # colorbar ----------------------------------------------------------------------------
-            subplot_cb = plt.subplot(gs[1])
+            cax = plt.subplot(gs[1])
             cbar_ticks = np.linspace(levels.min(), levels.max(), 11)
-            if cbar_ticks.max() == 1.0:
```
ksunden: I'm not sure I like this particular way of determining it, but do we wish to maintain single decimal place for normalized data?

I think it may be something we consider doing in a smarter way in general, not sure exactly how, but I feel that there should exist such a smart detection algorithm, that also scales to different orders of magnitude.

Unclear to me whether this detection should be moved to plot_colorbar or not

Other than that, it seems to function as expected from my limited testing.

PR #281, opened 2017-09-18
Hotfix for issue 279

wright-stuff:
Only a hotfix, doesn't resolve the issue, but will allow fitting to work for the time being.


untzag:
issue is fixed by #282

PR #301, opened 2017-09-22
add parse_ordered_strings

untzag:
I'm thinking that this can replace all of our calls parsing arguments of `channel`, `axis` etc


untzag:
forcing merge

Issue #304, opened 2017-09-26
data.from_KENT incorrectly assigns w1 and w2 to columns 1 and 2


neffmallon:
The correct columns are 2 and 1 respectively.

I know, I know. This is why we have headers now.

untzag:
wontfix

PR #305, opened 2017-09-26
Modifies data.from_KENT to resolve issue304

neffmallon:
#304


untzag:
@neffmallon is it true?

for example I'm fairly sure that the data in [this classic example](http://wright.tools/en/master/auto_examples/simple_2D.html#sphx-glr-auto-examples-simple-2d-py) imports as intended

if the OPA roles are inconsistent we should go about this in a different way

PR #314, opened 2017-09-29
artist updates

untzag:


WrightTools/artists.py
```diff
@@ -154,6 +338,117 @@ def legend(self, *args, **kwargs):
...
+        ylabel : string (optional)
+            ylabel. Default is None.
+        **kwargs
+            matplotlib.axes.Axes.pcolor__ optional keyword arguments.
+
+        __ https://matplotlib.org/api/_as_gen/matplotlib.axes.Axes.pcolor.html
+
+        Returns
+        -------
+        QuadContourSet
```
ksunden: Pretty sure pcolor doesn't return QuadContourSet

On all of these, they should be `matplotlib.collections.Collection` or similar to enable intersphinx

untzag: great catch

untzag:
thanks for keeping me honest @ksunden

PR #323, opened 2017-10-05
2.13.9

untzag:


untzag:
@ksunden I think the log is so big because you squashed?

ksunden:
I believe so too, I won't squash this time, and hopefully it reconciles them well from here on out

PR #339, opened 2017-10-12
join takes set of points along each axis (resolves #332)

### LINKED ISSUE #332: [before h5] join takes set of points along each axis
currently join looks at each axis and tries to guess what the best evenly spaced points are

moving forward, join will simply take the set of all points in all data objects along each axis

interpolation will still be used for points that are not contained in any data set (within the convex hull, of course)

untzag:


untzag:
while there is more to do with `join`, I think that it is prudent to wait until after h5

I've created #340 towards this end

as far as I'm concerned, this is ready for merge

PR #373, opened 2017-11-13
Split h5

ksunden:
Passes all tests on my local machine (Only had to comment out one line regarding constants from previous tests)

Things I'm not confident about:

* Whether this is handling 'parent' the way it is expected, I feel like we may wish to pass `edit_local=True` in that case
* Constants handling has been commented out for the time being, as it is not implemented upstream


WrightTools/data/_data.py
```diff
@@ -1834,65 +1834,97 @@ def split(self, axis, positions, units='same',
...
+                    attrs = dict(ch.attrs)
+                    attrs.pop('name', None)
+                    attrs.pop('units', None)
+                    new_data.create_channel(ch.natural_name, ch[:][slc], ch.units, **attrs)
+            else:
+                attrs = dict(self.attrs)
+                attrs.pop('name', None)
+                attrs.pop('axis_names', None)
+                attrs.pop('channel_names', None)
+                attrs.pop('constant_names', None)
```
untzag: I agree, we should pop `axis_names`, `channel_names` and `constant_names` within `Data.__init__` (and similar)

untzag:
please remove blank lines within methods (I know, it's a small point)

untzag:
thanks @ksunden

Issue #399, opened 2017-12-07
Provide null support for contour overload when using arrays


p770193:
When passing a `data` object, `.artists._parse_limits` uses channel attributes to refine it's limits calculations.  Similar calculations should be available with kwargs when passing an array.

ksunden:
I believe that this can be closed.
If there any objections @untzag @p770193, please reopen.

Issue #398, opened 2017-12-07
Overloaded contour commands misuse levels kwarg when data object not used


p770193:
When handing `contourf`/`contour` an array, e.g.
`
d1 = wt.data.from_PyCMDS(p)
d1.normalize()
plt.contourf(d1.w2.points, d1.d2.points, d1.signal_ratio.values.T,
             signed=True,
             levels=np.linspace(-1,1))
`
The plot will not "center" the colorbar so zero signal is white, but instead plot according to dynamic range.

ksunden:
I believe this can be closed.
If there are objections, please reopen @p770193 @untzag .

PR #401, opened 2017-12-07
_parse_limits, _parse_cmap handle zi better (3.0)

p770193:
Smarter determination of limits and cmap when handing plotting commands using an array instead of data object. Resolves issues #399 and #398


ksunden:
See my comments on #402  
Also, @p770193 please close the PR that you do not wish to be reviewed

PR #402, opened 2017-12-07
_parse_limits, _parse_cmap handle zi better (2.X)

p770193:
Smarter determination of limits and cmap when handing plotting commands using an array instead of data object. Resolves issues #399 and #398

### DIFF
--- a/WrightTools/artists.py
+++ b/WrightTools/artists.py
@@ -61,6 +61,9 @@ class Axes(matplotlib.axes.Axes):
             if isinstance(kwargs['cmap'], string_type):
                 kwargs['cmap'] = colormaps[kwargs['cmap']]
             return kwargs
+        elif 'signed' in kwargs.keys():
+            kwargs['cmap'] = colormaps['signed']
+            return kwargs
         if data:
             if data.channels[channel_index].signed:
                 kwargs['cmap'] = colormaps['signed']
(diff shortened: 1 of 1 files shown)


WrightTools/artists.py
```diff
@@ -61,6 +61,9 @@ def _parse_cmap(self, data=None, channel_index=None, **kwargs):
             if isinstance(kwargs['cmap'], string_type):
                 kwargs['cmap'] = colormaps[kwargs['cmap']]
             return kwargs
+        elif 'signed' in kwargs.keys():
+            kwargs['cmap'] = colormaps['signed']
+            return kwargs
```
ksunden: This can be accomplished with `cmap='signed'`
I'm hesitant to add behavior that is solely dependent on the presence of a kwarg, I feel like this could be confusing, e.g. `signed = False` here, would result in a signed colormap being used. If we wish to treat signed as a 'special' kwarg, I think we should look at the value.

WrightTools/artists.py
```diff
@@ -103,9 +106,20 @@ def _apply_labels(self, autolabel='none', xlabel=None, ylabel=None, data=None, c
             self.set_ylabel(ylabel, fontsize=18)
 
     def _parse_limits(self, zi=None, data=None, channel_index=None, dynamic_range=False, **kwargs):
+        null = kwargs.get('null', 0)
         if zi is not None:
-            vmin = np.nanmin(zi)
-            vmax = np.nanmax(zi)
+            if 'levels' in kwargs.keys():
```
ksunden: should this only be respected if `zi` is not None?

WrightTools/artists.py
```diff
@@ -103,9 +106,20 @@ def _apply_labels(self, autolabel='none', xlabel=None, ylabel=None, data=None, c
...
+                levels = kwargs['levels']
+                vmin = levels.min()
+                vmax = levels.max()
+            elif 'signed' in kwargs.keys():
+                if kwargs['signed'] and not dynamic_range:
+                    mag = max(np.abs([np.nanmin(zi-null), np.nanmax(zi-null)]))
+                    vmin = null - mag
+                    vmax = null + mag
```
ksunden: What about the case where dynamic range is True?

WrightTools/artists.py
p770193: Agreed, but more fundamental issues with `signed` kwarg have been raised.  Decided to remove the kwarg altogether (346b527f620f45109ccd3d0035e016ecdfbf0f06)

WrightTools/artists.py
p770193: Again, removed because `signed` kwarg is not desired (346b527f620f45109ccd3d0035e016ecdfbf0f06)

WrightTools/artists.py
p770193: Probably?  Can you offer your more informed opinion @untzag @ksunden about how `levels` should be treated when `data is not None`?

PR #414, opened 2017-12-18
data manipulation

untzag:
recovers the following methods of data:
- `bring_to_front`
- `remove_channel`
- `rename_attrs` (changed to `rename`, method of `Group`)

### DIFF
--- a/WrightTools/_base.py
+++ b/WrightTools/_base.py
@@ -280,7 +280,7 @@ class Group(h5py.Group):
         """
         # TODO: what happens to lists channel_names, variable_names etc?
         # maybe they are regenerated from scratch?
-        # 
+        #
         # ensure that items will remain unique
         changed = kwargs.keys()
         for k, v in kwargs.items():
--- a/WrightTools/_base.py
+++ b/WrightTools/_base.py
@@ -266,36 +266,3 @@ class Group(h5py.Group):
     def flush(self):
         """Ensure contents are written to file."""
         self.file.flush()
-
-    def rename(self, **kwargs):
-        """Rename a set of attributes.
-
-        The name will be set to str(val), and its natural naming identifier
-        will be wt.kit.string2identifier(str(val))
-
-        Keyword Arguments
-        -----------------
-        Each argument should have the key of a current axis or channel,
-            and a value which is a string of its new name.
-        """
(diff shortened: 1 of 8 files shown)


WrightTools/data/_data.py
```diff
@@ -1565,6 +1577,109 @@ def remove_channel(self, channel, verbose=True):
         if verbose:
             print('channel {0} removed'.format(name))
 
+    def remove_variable(self, variable, verbose=True):
```
ksunden: Do we wish to have both singular and plural remove variable?
If so, this should just call the plural version with implied=False, I think, but leaning towards just the one function

WrightTools/data/_data.py
```diff
@@ -1565,6 +1577,109 @@ def remove_channel(self, channel, verbose=True):
...
+            names[index] = v
+        self.variable_names = names
+        # update axes
+        new = self.axis_expressions
+        for i, v in enumerate(kwargs.keys()):
+            for j, n in enumerate(new):
+                new[j] = n.replace(v, '{%i}' % i)
+        for i, n in enumerate(new):
+            new[i] = n.format(*kwargs.values())
+        self.transform(new)
```
ksunden: Maintain units?

untzag: done

WrightTools/_base.py
```diff
@@ -191,6 +209,9 @@ def __version__(self):
             self.file.attrs['__version__'] = wt5_version
         return self.file.attrs['__version__']
 
+    def _update(self):
+        pass
```
ksunden: Does this serve a purpose?

WrightTools/_base.py
```diff
@@ -245,3 +266,36 @@ def close(self):
...
+
+        The name will be set to str(val), and its natural naming identifier
+        will be wt.kit.string2identifier(str(val))
+
+        Keyword Arguments
+        -----------------
+        Each argument should have the key of a current axis or channel,
+            and a value which is a string of its new name.
+        """
+        # TODO: what happens to lists channel_names, variable_names etc?
```
ksunden: Still todo?

WrightTools/_base.py
untzag: just as a placeholder

eventually want to migrate a bunch of natural naming code here

PR #430, opened 2018-01-05
copy and save

untzag:
`_update` becomes `_update_natural_namespace`, to signfiy what it axtually does

all `copy` and `save` methods now inherited from `Group` parent class

`save` now allows for saving nested groups as root of new file---to do this, `save` calls `copy`

simplify `parent`

add try, except to `close` to "fix" bug that I do not understand

### DIFF
--- a/WrightTools/_group.py
+++ b/WrightTools/_group.py
@@ -226,6 +226,7 @@ class Group(h5py.Group):
         if name is None:
             name = self.natural_name
         if parent is None:
+            from ._open import open as wt_open  # circular import
             new = Group()  # root of new tempfile
             # attrs
             new.attrs.update(self.attrs)
@@ -235,7 +236,6 @@ class Group(h5py.Group):
                 self[k].copy(new)
             new.flush()
             p = new.filepath
-            from ._open import open as wt_open
             new = wt_open(p)
         else:
             # copy
--- a/WrightTools/_group.py
+++ b/WrightTools/_group.py
@@ -139,8 +139,9 @@ class Group(h5py.Group):
             self.file.attrs['__version__'] = wt5_version
         return self.file.attrs['__version__']
 
-    def _update(self):
-        pass
+    def _update_natural_namespace(self):
+        for name in self.item_names:
+            setattr(self, name, self[name])
 
     @property
     def fullpath(self):
@@ -152,7 +153,7 @@ class Group(h5py.Group):
         """Item names."""
         if 'item_names' not in self.attrs.keys():
             self.attrs['item_names'] = np.array([], dtype='S')
-        return self.attrs['item_names']
+        return tuple(n.decode() for n in self.attrs['item_names'])
 
     @property
     def natural_name(self):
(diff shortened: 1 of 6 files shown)


WrightTools/_group.py
```diff
@@ -178,21 +179,113 @@ def parent(self):
...
+                self.file.flush()
+                self.file.close()
+            except SystemError:
+                pass
+            finally:
+                if hasattr(self, '_tmpfile'):
+                    os.close(self._tmpfile[0])
+                    os.remove(self._tmpfile[1])
+
+    def copy(self, parent=None, name=None, verbose=True):
```
ksunden: I'm slightly confused as to why we are not just relying on the underlying h5py `copy`
http://docs.h5py.org/en/latest/high/group.html#Group.copy

It seems to have most of what we want, except creation of a temp file. It recursively copies the content, including attributes

ksunden: natural name, as well, I suppose,

But, the inner workings I feel could be replaced by a call to `super` and adding the kwargs in that the underlying method supports, perhaps

WrightTools/_group.py
```diff
@@ -178,21 +179,113 @@ def parent(self):
...
+        # finish
+        if verbose:
+            print('{0} copied to {1}'.format(self.fullpath, new.fullpath))
+        return new
 
     def flush(self):
         """Ensure contents are written to file."""
         self.file.flush()
+
+    def save(self, filepath=None, verbose=True):
```
ksunden: I would like overwrite protection (i.e. `overwrite=False` kwarg, exception/warning when the save operation fails/destroys a previous copy)

WrightTools/_group.py
```diff
@@ -178,21 +179,113 @@ def parent(self):
...
+        new = Group(filepath=filepath)
+        # attrs
+        for k, v in self.attrs.items():
+            new.attrs[k] = v
+        # children
+        for k in self.keys():
+            self[k].copy(new, verbose=True)
+        # finish
+        import time
+        time.sleep(10)
```
ksunden: Why are we sleeping for ten seconds?????

If waiting is needed, please find a way to check how much wait is needed, rather than having a hard coded number.

Hard coded wait times always lose performance, very noticeably here, and either wait way too long on powerful machines, or not long enough on lower spec machines

Also, import in the middle of the function

WrightTools/_group.py
```diff
@@ -178,21 +179,113 @@ def parent(self):
...
+
+        Returns
+        -------
+        str
+            Written filepath.
+        """
+        # parse filepath
+        if filepath is None:
+            filepath = os.path.join(os.getcwd(), self.natural_name + '.wt5')
+        elif len(os.path.basename(filepath).split('.')) == 1:
```
ksunden: Do we wish to have a more robust check here?
I would tend to use other characters myself, but I can imagine a scenario where someone wants a dot in their file name, but still wants the extension added. e.g. saving a sequence of related files (single collection may be preferred within the WT ecosystem, but if interacting with other things...) as `name.000.wt5` `name.001.wt5` etc.

Perhaps we check that the last thing after the `.` is not in a list of known hdf5 subtype file extensions: (e.g. h5 hdf5 wt5 netcdf nc) etc.

I think it may be easiest to not try and cover them all, but cover the bare hdf5 extensions and wt5 to check that they are already there, otherwise add the wt5 extension.

Obviously users have the ability to change the extension externally, but still, it can be confusing.

tests/group/save.py
```diff
@@ -0,0 +1,76 @@
...
+
+here = os.path.abspath(os.path.dirname(__file__))
+
+
+# --- helpers -------------------------------------------------------------------------------------
+
+
+def assert_equal(a, b):
+    if hasattr(a, '__iter__'):
+        for ai, bi in zip(a, b):
```
ksunden: should check for equal length, perhaps as well

darienmorrow:
Ensure `_update_natural_namespace()` is called.

PR #442, opened 2018-01-09
Move flush outside of loop

ksunden:
Test still passed on my local machine, reduces overhead of chop operation by a factor of 2


untzag:
great insight, @ksunden

Issue #309, opened 2017-09-29
data.normalize limit kwargs


untzag:
I often find myself wanting to normalize data to some minor feature

sometimes it is a single peak in a large spectrum that contains much brighter features far from my region of interest

other times it's trying to visualize a feature in the center of a spectrum with very noisy edges

normalize could have kwargs to allow for normalization of some contiguous _region_ of the channel

it's not exactly clear how to parametrize it---one idea is to pass in 2-element list-likes with keys according to axis names

we should also propagate such syntax to `get_nadir` and `get_zenith`

we could also consider propagating such syntax to `clip`...

darienmorrow:
Are you interested in normalizing a region or normalizing to a region?

untzag:
@darienmorrow I don't fully understand your question

the entire channel will be scaled such that mag is one within the defined region

PR #443, opened 2018-01-10
operators

untzag:
- [x] `__iadd__`
- [x] `__imul__`
- [x] `__ipow__`
- [x] `__isub__`
- [x] `__itruediv__`
- [x] `symmetric_root`
- [x] remove `channel.invert`
- [x] `clip`
- [x] `log`, `log10`, `log2`
- [x] `channel.minor_extent`, `channel.major_extent`, `channel.mag`
- [x] `channel.normalize`


WrightTools/_dataset.py
```diff
@@ -25,8 +27,53 @@ class Dataset(h5py.Dataset):
...
+    def __isub__(self, value):
+        def f(dataset, s, value):
+            if hasattr(value, 'shape'):
+                dataset[s] -= value[wt_kit.valid_index(s, value.shape)]
+            else:
+                dataset[s] -= value
+        self.chunkwise(f, value=value)
+        return self
+
+    def __itruediv__(self, value):
```
ksunden: Should we add `__ifloordiv__` and `__imod__` for completeness sake

WrightTools/data/_channel.py
```diff
@@ -113,59 +104,17 @@ def signed(self):
...
-    def invert(self):
-        """Invert channel values."""
-        self[:] *= -1
-
     def mag(self):
         """Channel magnitude (maximum deviation from null)."""
         return self.major_extent
 
-    def normalize(self, axis=None):
+    def normalize(self):
```
ksunden: What if we wish to normalize along a particular axis?

tests/data/channel/normalize.py
```diff
@@ -0,0 +1,37 @@
...
+
+def test_wigner():
+    p = datasets.COLORS.v2p2_WL_wigner
+    data = wt.data.from_COLORS(p)
+    data.ai0.normalize()
+    assert data.ai0.null == 0.
+    assert data.ai0.max() == 1.
+    data.close()
+
+
```
ksunden: Test for signed data normalization?

WrightTools/data/_channel.py
```diff
@@ -91,17 +93,6 @@ def null(self):
             self.attrs['null'] = 0
         return self.attrs['null']
 
-    @property
-    def info(self):
```
ksunden: Any particular reason to remove the info dictionary?

WrightTools/_dataset.py
untzag: my vote is to add them later, only if we can think of some good reason for their existance

WrightTools/data/_channel.py
untzag: yeah, I just wasn't seeing it get used

do you think it's useful?

WrightTools/data/_channel.py
untzag: yeah, I was disapointed to do this, but I think it just needs to be thought about

I'll open an issue

tests/data/channel/normalize.py
untzag: yeah OK

believe it or not, I don't think I have any signed datasets in WrightTools

WrightTools/data/_channel.py
ksunden: Perhaps add it to #309

tests/data/channel/normalize.py
untzag: actually, I'm not sure what behavior I want for signed datasets

PR #448, opened 2018-01-12
units and labels

### LINKED ISSUE #416: api for "forcing" unit conversion of underlying variables
currently, `data.convert` and `axis.convert` simply change the units that are returned when slicing into the axis object---the variables themselves remain untouched---I like this behavior and want to keep it

in addition to the above, users should be able to convert the variables directly

step one: add `convert` method to `Variable` class

step two: add `convert_variables=False` kwarg to `data.convert`, `axis.convert`, which toggles propagation to dependent variables

we should think carefully, because variables can appear in multiple axes

### LINKED ISSUE #433: wt.artists.quick2D (and maybe 1D) don't plot  label seeds.
Suppose I have a data object `d` s.t. the attribute `d.d2.label` is `'$\\mathsf{\\tau_{21}\\,\\left(ps\\right)}$'`. In other words, the label has tau_21 noted on it. If I hand this data object to `quick2D` the d2 axis label is merely `'$\\mathsf{\\tau\\,\\left(ps\\right)}$'`. In other words, the label has no tau_21 on it. 

I have a hunch that data.chop is to blame.

untzag:
resolves #416 
resolves #433

### DIFF
--- a/WrightTools/_dataset.py
+++ b/WrightTools/_dataset.py
@@ -30,7 +30,6 @@ class Dataset(h5py.Dataset):
         if not hasattr(index, '__iter__'):
             index = [index]
         index = wt_kit.valid_index(index, self.shape)
-        print(index)
         return super().__getitem__(index)
 
     def __iadd__(self, value):
--- a/WrightTools/_dataset.py
+++ b/WrightTools/_dataset.py
@@ -6,7 +6,6 @@
 
 import posixpath
 import collections
-from concurrent.futures import ThreadPoolExecutor
 
 import numpy as np
 
@@ -212,8 +211,10 @@ class Dataset(h5py.Dataset):
             raise wt_exceptions.UnitsError(valid, destination_units)
         if self.units is None:
             return
-        def f(dataset, s, destination_units):
+       
+       def f(dataset, s, destination_units):
             dataset[s] = wt_units.converter(dataset[s], dataset.units, destination_units)
+        
         self.chunkwise(f, destination_units=destination_units)
         self.units = destination_units
 
--- a/WrightTools/_dataset.py
+++ b/WrightTools/_dataset.py
@@ -12,7 +12,9 @@ import numpy as np
 
 import h5py
 
+from . import exceptions as wt_exceptions
 from . import kit as wt_kit
+from . import units as wt_units
 
 
 # --- class ---------------------------------------------------------------------------------------
@@ -28,6 +30,7 @@ class Dataset(h5py.Dataset):
         if not hasattr(index, '__iter__'):
             index = [index]
         index = wt_kit.valid_index(index, self.shape)
+        print(index)
         return super().__getitem__(index)
 
     def __iadd__(self, value):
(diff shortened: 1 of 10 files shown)


WrightTools/_dataset.py
```diff
@@ -212,8 +211,10 @@ def convert(self, destination_units):
             raise wt_exceptions.UnitsError(valid, destination_units)
         if self.units is None:
             return
-        def f(dataset, s, destination_units):
+       
+       def f(dataset, s, destination_units):
```
ksunden: needs another space

WrightTools/kit/_array.py
```diff
@@ -270,11 +270,10 @@ def valid_index(index, shape):
     tuple
         Valid index.
     """
+    # TODO: fails in case of shape (25, 1), index slice(None)
```
ksunden: Is this still true? Either remove TODO or open an issue

WrightTools/units.py
```diff
@@ -4,34 +4,31 @@
...
-         'fs': ['x', 'x', r'fs'],
+delay = {'fs': ['x', 'x', r'fs'],
          'ps': ['x*1e3', 'x/1e3', r'ps'],
          'ns': ['x*1e6', 'x/1e6', r'ns'],
          'mm_delay': ['x*2*fs_per_mm', 'x/(2*fs_per_mm)', r'mm']}
 
 # energy units (native: nm)
-energy = {'kind': 'energy',
-          'nm': ['x', 'x', r'nm'],
+energy = {'nm': ['x', 'x', r'nm'],
```
ksunden: I'm mildly inclined to make the 'native' units for energy an actual energy unit (perhaps wn?)
Scaler multiples just seems cleaner to me (inverse still needed for nm, of course)

Is now the time to consider this change?

WrightTools/units.py
```diff
@@ -111,86 +113,71 @@ def converter(val, current_unit, destination_unit):
...
+        d['GHz'] = r'f'
+        return d.get(units, 'E')
+    elif kind(units) == 'delay':
+        return r'\tau'
+    elif kind(units) == 'fluence':
+        return r'\mathcal{F}'
+    elif kind(units) == 'pulse_width':
+        return r'\sigma'
+    else:
+        return ''
```
ksunden: Should the default be the unit itself?

WrightTools/kit/_array.py
untzag: yes, this is actually a big problem---embarising that I forgot about it

don't merge for now

WrightTools/units.py
untzag: sure, that's fine

WrightTools/units.py
untzag: :thinking: 

probably not? because the unit appears elsewhere in the label

actually, the kind itself might be a good default

WrightTools/units.py
untzag: actually, let's skip it for now

I would accept a PR for it though :wink:

ksunden: I'll open an issue, mark as help wanted

tests/kit/valid_index.py
```diff
@@ -0,0 +1,58 @@
...
+    shape = (13, 25, 99)
+    assert wt.kit.valid_index(index, shape) == (slice(None), slice(None), slice(None))
+
+
+def test__s_s__51():
+    index = (slice(None), slice(None))
+    shape = (51,)
+    assert wt.kit.valid_index(index, shape) == (slice(None),)
+
+
```
ksunden: Test for slice which is not `slice(None)`

tests/kit/valid_index.py
```diff
@@ -0,0 +1,58 @@
...
+def test__s__13_25_99():
+    index = (slice(None),)
+    shape = (13, 25, 99)
+    assert wt.kit.valid_index(index, shape) == (slice(None), slice(None), slice(None))
+
+
+def test__s_s__51():
+    index = (slice(None), slice(None))
+    shape = (51,)
+    assert wt.kit.valid_index(index, shape) == (slice(None),)
```
ksunden: Is this expected behavior? Unclear to me that dropping a "slice" is appropriate, particularly for slices which are not `slice(None)`. Wonder if an exception should be thrown.

tests/kit/valid_index.py
ksunden: The way the code is written, It does not change any slices, and assumes they are valid slices...
It's unclear to me whether this should be how it behaves or not, I wonder if perhaps it should be looking at the slices and modifying or raising exceptions if those slices are not valid.

WrightTools/units.py
ksunden: this is #449

PR #450, opened 2018-01-15
print tree

untzag:
print a "pretty" tree representation of the contents of a `Collection`

### DIFF
--- a/WrightTools/_dataset.py
+++ b/WrightTools/_dataset.py
@@ -101,6 +101,14 @@ class Dataset(h5py.Dataset):
             del self.attrs['min']
 
     @property
+    def _leaf(self):
+        out = self.natural_name
+        if self.units is not None:
+            out += ' ({0})'.format(self.units)
+        out += ' {0}'.format(self.shape)
+        return out
+
+    @property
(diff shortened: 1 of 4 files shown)


ksunden:
I did something similar in `h5cli`, https://github.com/ksunden/h5cli/pull/60/

I used https://pypi.python.org/pypi/tree-format to get the formatting done

ksunden:
Some oddities when playing around with this, some of which are not (directly) related to the code here:

1. `col._items` is apparently not updated when `copy` is used to populate a collection. (`items()` and `item_names` are updated, however)
    This results in data objects which are copied in being left out of the tree representation, and the indexes being off.

2. I wonder if it might be useful to include variables and channels from data objects as part of the tree view. (and, perhaps be able to call `print_tree` on a data object itself) This would provide a quick overview of the shape of each constituent component, as well as the whole. Also, it would give a good idea of which variables are available for `transform`

untzag:
@ksunden I think we should open a separate issue for `_items_` propagation

ksunden:
Tested locally because Travis backlog forced them to jettison the queue.

PR #454, opened 2018-01-26
remove _items

untzag:
the idea of keeping an `_items` attribute was partially formed, and caused some strange problems


untzag:
travis errored... restarting build

PR #455, opened 2018-01-27
[WIP] instantiation

untzag:
improvements to instantiation

### DIFF
--- a/WrightTools/_group.py
+++ b/WrightTools/_group.py
@@ -25,7 +25,14 @@ wt5_version = '0.0.0'
 # --- class ---------------------------------------------------------------------------------------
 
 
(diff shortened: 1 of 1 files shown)


untzag:
note that I feel like `__new__` and `__init__` could be further streamlined now that we have proper access to `MetaClass.__call__`

however in the interest of quickly getting a working API, I think "working as intended" is a reasonable goal

untzag:
if the reviewer can think of any additional tests, let me know

ksunden:
I agree that there may be some streamlining we can do, but this is a good first step.

As for tests, Not sure there is much to explicitly test on its own, the behavior is more baked into how we instantiate objects in all the other tests.

Issue #475, opened 2018-02-11
recover subtract


ksunden:

ksunden:
To whoever implements this (may be myself, but we will see): Strong suggestion to make this a function which is then a single line in each dunder method.

PR #488, opened 2018-02-14
Remove Subtract/Divide, add base exception classes

ksunden:
See #478 #475


ksunden:
This is only failing because of the sphinx versions

PR #494, opened 2018-02-24
Add from_directory function

### LINKED ISSUE #493: Instantiating `Data` or `Collection` object with `parent` kwarg
```
>>> col = wt.Collection()
>>> col.create_data('hello')
<WrightTools.Data 'hello' () at /tmp/khcldndt.wt5::/hello>
>>> col.print_tree()
collection (/tmp/khcldndt.wt5)
└── 0: hello ()
    ├── axes: 
    └── channels: 
>>> wt.data.Data(parent=col)
<WrightTools.Data 'data' () at /tmp/khcldndt.wt5::/data>
>>> wt.data.Data(name='world', parent=col)
<WrightTools.Data 'world' () at /tmp/khcldndt.wt5::/world>
>>> col.print_tree()
collection (/tmp/khcldndt.wt5)
└── 0: hello ()
    ├── axes: 
    └── channels: 
>>> 
```

Note that the data object does indeed appear in the same file, in the location it should, however the parent does not recognize the object created with `wt.data.Data` in its `items`

The Data object is still accessible using `col['data']`

It can be added to the items list using `col.create_data('data')` and will retain all other modifications.

ksunden:
Note, It's not perfect yet, in part due to #493 
In particular, `from_Cary` uses the `Collection` syntax instead of `create_collection` (as the parent may be `None`) 

resolves #493 

Also need tests.


WrightTools/collection/_directory.py
```diff
@@ -0,0 +1,90 @@
...
+
+    for i in os.listdir(filepath):
+        q.put((filepath, i, root))
+
+    while not q.empty():
+        path, fname, parent = q.get()
+        for pattern, func in from_methods.items():
+            if fnmatch.fnmatch(fname, pattern):
+                if func is not None:
+                    func(os.path.join(path, fname), name=os.path.splitext(fname)[0], parent=parent,
```
untzag: why pass in `name`? shouldn't the from method get to decide what the best data object name is?

I guess using this strategy you are guaranteeing unique names

(non-blocking comment)

tests/collection/from_directory.py
```diff
@@ -0,0 +1,27 @@
...
+                 'COLORS':None,
+                }
+    p = wt.datasets.PyCMDS.w1_000
+    p = os.path.dirname(p)
+    p = os.path.dirname(p)
+    col = wt.collection.from_directory(p, from_dict)
+    # Print the tree in case tests fail, also will catch errors in printing
+    col.print_tree()
+    assert col.natural_name == 'datasets'
+    assert col.PyCMDS['w1 000']
```
untzag: are we getting invalid identifiers as names? seems reasonable to coerce to identifier, since natural naming is so core to the "WrightTools experience"

I consider this a bug right now, but am happy to be convinced otherwise

WrightTools/collection/_directory.py
ksunden: Failing to pass name causes the following exception:

```
  File "<input>", line 1, in <module>
    col = wt.collection.from_directory('.', {'KENT':None, 'COLORS':None, '*.data
':wt.data.from_PyCMDS, '*.csv':wt.collection.from_Cary, '*.dpt':wt.data.from_Ten
sor27}, verbose=True)
  File "/home/kyle/wright/wrighttools/WrightTools/collection/_directory.py", lin
e 81, in from_directory
    verbose=verbose)
  File "/home/kyle/wright/wrighttools/WrightTools/data/_pycmds.py", line 88, in 
from_PyCMDS
    data.create_variable(name='labtime', values=values)
  File "/home/kyle/wright/wrighttools/WrightTools/data/_data.py", line 525, in c
reate_variable
    id = self.require_dataset(name=name, data=values, shape=shape, dtype=dtype).
id
  File "/usr/lib/python3.6/site-packages/h5py/_hl/group.py", line 134, in requir
e_dataset
    raise TypeError("Shapes do not match (existing %s vs new %s)" % (dset.shape,
 shape))
TypeError: Shapes do not match (existing (51,) vs new (81, 81))
```

I'm not sure right now exactly why, but suspect it does indeed have to do with ensuring unique names.

My goal was to replicate the file structure as closely as possible.

ksunden: My personal opinion is that this method is designed for use in cases such as publication figure generation scripts, where the user may wish to override the 'name' field as stored in a `.data` file for instance. I would rather encourage changing to informative data file names rather than encouraging changing of the contents of the file. I also don't wish to require users to manually rename things after calling this.

tests/collection/from_directory.py
ksunden: My goal was to replicate the file structure as closely as possible, without messing with names if I could avoid it. I personally tend to have file names which are valid identifiers (without the extension, that is). I do this in part so this is not an issue in the things I do.

I'll note, prior behavior, without passing the name to the from method, does not guarantee a python identifier either: I have datasets which have the automatically generated name start with a numeral.

ksunden: To kind of sum up my thoughts: I don't think we should dictate what the names can or can't be. Having natural namespace is, to me, a useful thing that I wish to utilize, but it is a nicety, not a requirement. I do not want to build in the assumption that all file names will be valid python identifiers, in part because I do not want to lock the format specification to python

PR #513, opened 2018-02-27
Darien's little fixes. 

darienmorrow:
Will note that memory warnings are untested because the [warning just passes](https://github.com/wright-group/WrightTools/blob/development/WrightTools/exceptions.py#L102).


darienmorrow:
I'm still waiting on pep8speaks to yell at me...

untzag:
@darienmorrow pep8speaks only talks if there are violations

PR #515, opened 2018-02-27
Travis building with pip

ksunden:
Removes testing for 3.4, actually builds on 3.5 and 3.6


setup.py
```diff
@@ -32,7 +32,7 @@ def package_files(directory):
     packages=find_packages(),
     package_data={'': extra_files},
     setup_requires=['pytest-runner'],
-    tests_require=['pytest', 'pytest-cov'],
+    tests_require=['setuptools==38.5.1', 'pytest', 'pytest-cov', 'sphinx==1.6.5', 'sphinx-gallery==0.1.12', 'sphinx-rtd-theme'],
```
untzag: pep8

untzag:
so there are two versions of WrightTools in the container---the cloned one and the installed one, right?

are we sure we are actually importing the installed one? if we are in the cloned directory, I worry that we might not be

PR #534, opened 2018-03-06
Remove nans additional test and reimplementation

### LINKED ISSUE #532: remove_nans_1D fails for list
```
>>> wt.kit.remove_nans_1D([np.nan, 1, 2, 2])
Traceback (most recent call last):
  File "<input>", line 1, in <module>
    wt.kit.remove_nans_1D([np.nan, 1, 2, 2])
  File "/home/kyle/wright/WrightTools/WrightTools/kit/_array.py", line 223, in rem
ove_nans_1D
    return tuple(a[goods] for a in args)
  File "/home/kyle/wright/WrightTools/WrightTools/kit/_array.py", line 223, in <ge
nexpr>
    return tuple(a[goods] for a in args)
TypeError: list indices must be integers or slices, not list
>>> wt.kit.remove_nans_1D(np.array([np.nan, 1, 2, 2]))
(array([1., 2., 2.]),)
```

ksunden:
Closes #532 
 
(This was causing a problem within fit, which was run during the documentation building)

### DIFF
--- a/WrightTools/kit/_array.py
+++ b/WrightTools/kit/_array.py
@@ -210,17 +210,10 @@ def remove_nans_1D(*args):
     tuple
         Tuple of 1D arrays in same order as given, with nan indicies removed.
     """
-    # find all indicies to keep
-    bads = np.array([])
-    for arr in args:
-        bad = np.array(np.where(np.isnan(arr))).flatten()
-        bads = np.hstack((bad, bads))
-    if hasattr(args, 'shape') and len(args.shape) == 1:
-        goods = [i for i in np.arange(args.shape[0]) if i not in bads]
-    else:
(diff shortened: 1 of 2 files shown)


ksunden:
```
>>> a = np.array([np.nan, 1, 2, 2])
>>> np.isnan(a)
array([ True, False, False, False])
>>> not np.isnan(a)
Traceback (most recent call last):
  File "<input>", line 1, in <module>
    not np.isnan(a)
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any()
 or a.all()
>>> np.isnan(a) == False
array([False,  True,  True,  True])
>>> np.isnan(a) is False
False
```
Hush, pep8speaks, that doesn't work here....

ksunden:
actually, `~a` works in the above example

PR #537, opened 2018-03-10
fix discover dimensions

untzag:
:bug: 
- [x] update code
- [x] tests


untzag:
to the reviewer, I've manually confirmed that the two datasets I removed have missing slices

not sure how they got to be missing, but they are certainly not pristine examples of COLORS and KENT data

since they are not uniquely providing references for their particular data format, I decided to remove them

PR #543, opened 2018-03-13
allow for constants in variables

### LINKED ISSUE #540: multiplication doesn't seem to work in d.transform
`d.transform('w2+w2+w2', 'w1', 'd2')` works.

`d.transform('3*w2', 'w1', 'd2')` does not work (even with varying spacing around the multiplication operator)
Part of the error that is raised:
```
  File "/home/darien/source/WrightTools/WrightTools/data/_data.py", line 1306, in transform
    axis = current.get(expression, Axis(self, expression))

  File "/home/darien/source/WrightTools/WrightTools/data/_axis.py", line 53, in __init__
    self.units = self.variables[0].units

  File "/home/darien/source/WrightTools/WrightTools/data/_axis.py", line 152, in variables
    return self._variables

AttributeError: 'Axis' object has no attribute '_variables'
```

untzag:
resolves #540 
- [x] fix basic issue
- [x] consider what should be done about invalid axis identifiers (such as `3__t__w1`), if anything
- [x] write tests

### DIFF
--- a/WrightTools/data/_axis.py
+++ b/WrightTools/data/_axis.py
@@ -146,7 +146,10 @@ class Axis(object):
         except (AssertionError, AttributeError):
             pattern = '|'.join(map(re.escape, operators))
             keys = re.split(pattern, self.expression)
-            indices = [self.parent.variable_names.index(key) for key in keys]
+            indices = []
+            for key in keys:
+                if key in self.parent.variable_names:
+                    indices.append(self.parent.variable_names.index(key))
             self._variables = [self.parent.variables[i] for i in indices]
         finally:
             return self._variables
--- a/WrightTools/data/_axis.py
+++ b/WrightTools/data/_axis.py
@@ -106,7 +106,7 @@ class Axis(object):
         name = self.expression.strip()
         for op in operators:
             name = name.replace(op, operator_to_identifier[op])
-        return name
+        return wt_kit.string2identifier(name)
 
     @property
     def ndim(self):
--- a/tests/data/transform.py
+++ b/tests/data/transform.py
@@ -4,6 +4,8 @@
 # --- import --------------------------------------------------------------------------------------
 
 
+import numpy as np
+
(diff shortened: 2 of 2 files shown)


untzag:
also consider that floats appearing will be a disaster for natural naming `0.5__t__w1` is invalid for multiple reasons

ksunden:
Yes, floats are a problem for `__getattr__` (and we already subsumed `__d__` (for "dot") and `__p__` (for "point"))

At that point, you simply must use `data.axes[index]`, but I do feel that floats should be something axes can handle. The `getattr` problems aside, they are trivial to support in actual function since we use `numexpr` for evaluation.

ksunden:
I think for the simple number case, prepending `_` is the cleanest way to get around the issue. This is similar to what we do in `string2identifier`. Sure, it hides it from autosuggest, but I don't see a better way around it, and I think it is more natural to think with the numeral coming first  (including in what goes on in axis labels and the like where the expression is used)

WrightTools/data/_axis.py
```diff
@@ -146,7 +146,10 @@ def variables(self):
         except (AssertionError, AttributeError):
             pattern = '|'.join(map(re.escape, operators))
             keys = re.split(pattern, self.expression)
-            indices = [self.parent.variable_names.index(key) for key in keys]
+            indices = []
+            for key in keys:
+                if key in self.parent.variable_names:
+                    indices.append(self.parent.variable_names.index(key))
```
ksunden: Could still be done with lost comprehension

untzag: yeah but I think that this is more readable, to be honest---just a mild preference

I'll update to list comprehension if you want

ksunden: Fair enough, just bring it up as it's possible list comprehension maybe more performance (though these are cached anyway). Just something to think about/have in the back of our minds/on the record.

PR #559, opened 2018-03-19
level now works with more than zeroth channel. (resolves #558)

### LINKED ISSUE #558: level only works for zeroth channel
[L663](https://github.com/wright-group/WrightTools/blob/development/WrightTools/data/_data.py#L663) is the culprit, I think. 
`channel = self.channels[0]`
should instead be
`channel = self.channels[channel_index]`

darienmorrow:


ksunden:
Can we get a test case for this to ensure it does not pop up again?

Thanks for the PR, Darien.

PR #564, opened 2018-03-20
remove empty subscripts in labels

untzag:
currently there is bad behavior in axis labels which cause the following 'clipping' when plotted using mathjax

before 000

I finally understand that this happens because of the empty subscript in the generated label string: `$\mathsf{\bar\nu_{}\,\left(cm^{-1}\right)}$`

note that if LaTeX is used to render the text (`wt.artists.apply_rcparams('publication')`) the clipping does not occur---so in some sense this might be considered a bug in mathjax or matplotlib's implementation thereof

regardless of fault, it's an easy problem to fix on our end

this simple commit ensures that empty subscripts are not propagated, such that the generated label string is `$\mathsf{\bar\nu\,\left(cm^{-1}\right)}$`

this results in the following, correctly formatted, figure

after 000

### DIFF
--- a/WrightTools/data/_axis.py
+++ b/WrightTools/data/_axis.py
@@ -90,7 +90,9 @@ class Axis(object):
         symbol = wt_units.get_symbol(self.units)
         label = r'$\mathsf{' + self.expression
         for v in self.variables:
-            label = label.replace(v.natural_name, '%s_{%s}' % (symbol, v.label))
+            vl = '%s_{%s}' % (symbol, v.label)
+            vl = vl.rstrip('_{}')  # label can be empty, should not propagate trailing underscore
+            label = label.replace(v.natural_name, vl)
         if self.units_kind:
             units_dictionary = getattr(wt_units, self.units_kind)
             label += r'\,'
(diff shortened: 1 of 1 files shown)


untzag:
note that I think it's important to continue to allow for empty variable labels

I thought briefly about falling back to the natural name as subscript in cases where the label is empty, but I think it is a legitimate use case to have symbols without subscripts if the user so desires

WrightTools/data/_axis.py
```diff
@@ -90,7 +90,9 @@ def label(self):
         symbol = wt_units.get_symbol(self.units)
         label = r'$\mathsf{' + self.expression
         for v in self.variables:
-            label = label.replace(v.natural_name, '%s_{%s}' % (symbol, v.label))
+            vl = '%s_{%s}' % (symbol, v.label)
+            vl = vl.rstrip('_{}')  # label can be empty, should not propagate trailing underscore
```
ksunden: https://docs.python.org/3.6/library/stdtypes.html?highlight=rstrip#str.rstrip

ksunden: Rstrip will strip any character in the argument string, not just the full sequence

PR #572, opened 2018-03-25
plot colormap components

untzag:
fixing up an old friend...

generates figures like this
cubehelix

EDIT: added grayscale
viridis

### DIFF
--- a/WrightTools/artists/_colors.py
+++ b/WrightTools/artists/_colors.py
@@ -14,6 +14,7 @@ import matplotlib.colors as mplcolors
 import matplotlib.gridspec as grd
 
 
+
 # --- define -------------------------------------------------------------------------------------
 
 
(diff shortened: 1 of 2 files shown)


WrightTools/artists/_colors.py
```diff
@@ -191,28 +192,32 @@ def plot_colormap_components(cmap):
...
+    plt.plot(xi, g, 'g', linewidth=5, alpha=0.6)
+    plt.plot(xi, b, 'b', linewidth=5, alpha=0.6)
+    plt.plot(xi, k, 'k', linewidth=5, alpha=0.6)
+    ax.set_xlim(0, 1)
+    ax.set_ylim(-0.1, 1.1)
     # finish
-    plt.grid()
-    plt.xlabel('value', fontsize=17)
-    plt.ylabel('intensity', fontsize=17)
+    from ._helpers import set_ax_labels  # recursive import protection
```
ksunden: put at top of method

examples/plot_colormap_components.py
```diff
@@ -0,0 +1,12 @@
...
+Plot colormap components
+========================
+
+Quickly plot the RGB components of a colormap.
+"""
+
+import WrightTools as wt
+
+cmap = wt.artists.colormaps['default']
+wt.artists.plot_colormap_components(cmap)
```
ksunden: Would it be useful to have a comparison of several colormaps on the same page?

untzag: probably but it's not within the scope right now

note that we have a figure with all of the cmaps on the artists page

PR #586, opened 2018-04-19
use ast and re for string2array heavy lifting

ksunden:


darienmorrow:
@ksunden I checked-out this branch on the fs computer and ran the workup script which failed last night. Everything seems to be fine!

darienmorrow:
I have no idea the black magic you are doing, @ksunden, but you replaced 35 lines of code with 3. 💯

untzag:
so sorry to be so eternally irresponsible 

are we sure we need `literal_eval`?

was thinking of using `np.loadtxt` using the `io.StringIO` trick

but this would only work for up to 2 dimensions?

could loop

PR #594, opened 2018-05-16
Filters dataset and Cary fixes to make it work

ksunden:
There are still a few things that may be a bit weird about using '%T' as the channel name, but overall seems to work. Mostly it's just `getattr` dot syntax that does not work well as it is not a valid identifier, and does not get a valid identifier added by `setattr`


untzag:
@ksunden obviously we cannot change the .data files that are already written, but can we stop writing new files with a leading % in the column names? We can just move to "T" instead of "%T".

untzag:
oh gosh I'm thinking about the wrong thing---ignore my comment

Issue #587, opened 2018-04-23
quick2D fails with dimensionality error.


darienmorrow:
I have tried three ways to get quick2D to plot some 2D freqs. All have failed. One error message is:
```
  File "/home/darien/source/WrightTools/WrightTools/artists/_base.py", line 363, in pcolor
    raise wt_exceptions.DimensionalityError(2, data.ndim)

DimensionalityError: (DimensionalityError(...), 'dimensionality must be 2 (recieved 1)')
```
Note, quick2D works when I plot yaxis=1 and xaxis=0 (which plots ND wheel vs w2).

Minimum working example:
```
import WrightTools as wt
d = wt.data._pycmds.from_PyCMDS(r'early.data')
d.signal_ratio.signed = True
#wt.artists.quick2D(d, channel='signal_ratio', yaxis=1, xaxis=2, autosave=True)
#wt.artists.quick2D(d, channel='signal_ratio', yaxis='w2', xaxis='w1__e__wm', autosave=True)
wt.artists.quick2D(d, channel='signal_ratio', yaxis='w2', xaxis='w1=wm', autosave=True)
```
[early.zip](https://github.com/wright-group/WrightTools/files/1938037/early.zip)

Please rename issue with better name if fitting.

darienmorrow:
I think the problem is with `chop`.
`out = d.chop('w1=wm', 'w2')` yields a collection in which each data object has dimension (36,1), instead of the desired (36,36).

ksunden:
Chop has some issues with equality expressions, I believe (and perhaps expressions in general)

Can you try transforming the data object such that the axis is just wm or w1,not wm=w1?

ksunden:
The problem here is kind of two fold (and I've seen it before, though I thought I had fixed it, apparently not):

from_PyCMDS is creating 'd1' with a shape of 4x36x1, despite the fact that it is a scanned axis which should just have a shape of 4x1x1.
This is causing problems in chop (which Ideally chop would be able to handle itself.)

So, in summary, from_PyCMDS is doing something which is not intended, but shouldn't be illegal.
Chop is not set up to handle that kind of data at this time (and will require a fair bit more thinking to get there, though it is doable)

ksunden:
This was fixed by #596

Issue #553, opened 2018-03-17
more robust clip test


ksunden:
The current test for clip fails regularly, but stochastically. The function works, but is clipping all values

ksunden:
Perhaps could be as simple as explicitly setting the random seed in the test (though at that point why use random at all?)

PR #596, opened 2018-05-28
Tolerance, centers, and points

ksunden:
TODO: Write explicit test for this. 
Would also like explicit assertions in existing tests for centers and points being present and properly shaped.

@darienmorrow do you have a dataset that I can use which fails (to be properly read in, not necessarily failing to read in at all) on current development?

Also test agianst:

- [x] cobalamin paper
- [x] Darien active paper[s]
- [x] bearclaw paper [using existing file, and run from scratch]


ksunden:
failed test was due to #553 , rerunning

darienmorrow:
@ksunden I think #587 has some data that will fail `development` when it gets chopped. Can you download the data and see if it is what you need? If not, I can trudge through my archive.

darienmorrow:
Are the comments starting on [line 99](https://github.com/wright-group/WrightTools/blob/tolerance_inf/WrightTools/data/_pycmds.py#L99) necessary still? 

Can we provide some documentation in the docstring for how tolerances now work?

ksunden:
Breaks in `join` for certain datas (where we are joining different dimensioned scans

Should make `join` use the set of variables which are in all datas to be joined

ksunden:
Travis timed out, I'm not sure what happened, but it didn't _start_ the actual tests until around 45 minutes in, so

PR #602, opened 2018-06-02
migrate requirements

untzag:
from what I can see, we have been using `requirements.txt` wrong

[see this discussion on python.org](https://packaging.python.org/discussions/install-requires-vs-requirements/)

WrightTools' requirements should be in `setup.py` _first_, as this PR establishes

by having the stub requirements file, we still get the `pip install -r requirements.txt` behavior that we like

but distribution does not rely on requirements anymore

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,39 +11,61 @@ def read(fname):
     return open(os.path.join(here, fname)).read()
 
 
-extra_files = {'WrightTools': ['datasets', 'datasets/*', 'datasets/*/*', 'datasets/*/*/*',
-                               'datasets/*/*/*/*', 'VERSION', 'WT5_VERSION']}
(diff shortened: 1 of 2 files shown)


untzag:
@ksunden please pull and investigate before approving

setup.py
```diff
@@ -11,39 +11,61 @@ def read(fname):
...
+    setup_requires=["pytest-runner"],
+    tests_require=[
+        "pytest",
+        "pytest-cov",
+        "sphinx==1.6.5",
+        "sphinx-gallery==0.1.12",
+        "sphinx-rtd-theme",
+    ],
+    install_requires=[
+        "h5py",
```
ksunden: Should we still pin version to >=2.7

PR #615, opened 2018-06-10
more descriptive response when converting axes (issue #107)

### LINKED ISSUE #107: more descriptive response when converting axes
Would be nice to have source and destination units printed during conversion. This is especially useful when the conversion happens inside of some method.

Current behavior: `axis w1 converted`
Desired behavior: `axis w1 converted from wn to eV`

parkertomf:
Also added .format print and the corresponding changes to quickstart.

Closes #107


ksunden:
Closes #107

WrightTools/data/_data.py
```diff
@@ -427,16 +437,23 @@ def convert(self, destination_units, *, convert_variables=False, verbose=True):
...
             if axis.units_kind == units_kind:
-                axis.convert(destination_units, convert_variables=convert_variables)
                 if verbose:
-                    print("axis", axis.expression, "converted")
+                    print(
+                        "axis {} converted from {} to {}".format(
+                            axis.expression, axis.units, destination_units
+                        )
+                    )
+                axis.convert(destination_units, convert_variables=convert_variables)
```
untzag: @parkertomf can you move the call to `axis.convert` _before_ the `if verbose` switch?

that way the print statement is only made if the axis was actually converted successfully

ksunden: (note this applies to both lines here)

PR #637, opened 2018-06-18
Add warning and explicit return when existing object is created

### LINKED ISSUE #621: raise a warning when dupicate objects "created" by users
When you create two or more collections, `root.create_collection()` but fail to specify the `name` kwarg, and you assign each to a different variable (`var1` and `var2`), then unintentional behavior can result in the longterm e.g. trying to add data to the individual collections using from pycmds `wt.data.from_PyCMDS(p, parent=var1)` and `wt.data.from_PyCMDS(p, parent=var2)` will throw an error when run together if `var1` and `var2` are both collections created without specifying their names. 

As an end user this behavior is a bit wonky.
Note, this could be something to think about in create_data too...

ksunden:
closes #621

### DIFF
--- a/WrightTools/collection/_collection.py
+++ b/WrightTools/collection/_collection.py
@@ -7,6 +7,7 @@
 import numpy as np
 
 from .. import data as wt_data
+from .. import exceptions as wt_exceptions
 from .._group import Group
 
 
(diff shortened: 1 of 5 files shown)


ksunden:
build failed python 3.6 because it could not find imageio for some reason, rerunning

PR #638, opened 2018-06-19
axes methods return

untzag:
@wright-group/wrighttools-maintainers if you want me to add tests I will, but I need to think about the best way of doing so...

tempted to put off testing artists for another day


ksunden:
I agree, test another day

ksunden:
Approved, pending test completion, merge upon completion

PR #646, opened 2018-06-25
allow mag as a kwarg to normalize

untzag:
by default, normalize scales a channel such that `null` goes to zero and `mag` goes to one

adding kwarg such that users may specify a specific target for `mag`

- [x] basic behavior
- [x] consistency in all places normalize is offered
- [x] tests

### DIFF
--- a/WrightTools/data/_channel.py
+++ b/WrightTools/data/_channel.py
@@ -107,7 +107,7 @@ class Channel(Dataset):
         """Channel magnitude (maximum deviation from null)."""
         return self.major_extent
 
-    def normalize(self):
+    def normalize(self, factor=1):
         """Normalize a Channel, set `null` to 0 and the mag to 1."""
 
         def f(dataset, s, null, mag):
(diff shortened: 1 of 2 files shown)


WrightTools/data/_channel.py
```diff
@@ -107,17 +107,19 @@ def mag(self):
...
             dataset[s] /= mag
 
         if self.signed:
-            mag = self.mag()
+            mag = self.mag() / factor
         else:
-            mag = self.max()
+            mag = self.max() / factor
+
+        print(mag, "2@@@@@@@@@@@@@@@@@@@@@")
```
ksunden: extra print

tests/data/channel/normalize.py
```diff
@@ -20,6 +22,17 @@ def test_LDS821():
     data.close()
 
 
+def test_LDS821_mag_random():
+    p = datasets.BrunoldrRaman.LDS821_514nm_80mW
+    data = wt.data.from_BrunoldrRaman(p)
+    mag = random.random() + 0.5
+    data.signal.normalize(mag)
+    assert data.signal.null == 0.
+    assert data.signal.max() == mag
```
ksunden: dangerous to do float equality comparison

tests/data/channel/normalize.py
```diff
@@ -4,6 +4,8 @@
 # --- import --------------------------------------------------------------------------------------
 
 
+import random
```
ksunden: Not a huge fan of random in tests... Can cause stochastic, hard to track down problems (see clip)

PR #657, opened 2018-07-03
from_Solis

### LINKED ISSUE #656: `from_andor` method
generate `Data` object from ascii exports of CCD

p770193:
Method for importing Andor spectrometer CCD data (exported ascii format).  

resolves #656 

- [x] write from function
- [x] add example dataset(s), describe in docs
- [x] write test(s)

### DIFF
--- a/WrightTools/data/__init__.py
+++ b/WrightTools/data/__init__.py
@@ -7,7 +7,7 @@ from ._channel import *
 from ._join import *
 from ._variable import *
 
-from ._andor import *
+from ._solis import *
 from ._brunold import *
 from ._colors import *
 from ._data import *
--- a/WrightTools/data/__init__.py
+++ b/WrightTools/data/__init__.py
@@ -7,6 +7,7 @@ from ._channel import *
 from ._join import *
 from ._variable import *
 
+from ._andor import *
 from ._brunold import *
 from ._colors import *
 from ._data import *
@@ -26,6 +27,7 @@ __all__ = [
     "Channel",
     "Variable",
     # From methods in alphabetic order
+    "from_andor",
     "from_BrunoldrRaman",
     "from_COLORS",
     "from_JASCO",
--- a/WrightTools/data/__init__.py
+++ b/WrightTools/data/__init__.py
@@ -27,7 +27,6 @@ __all__ = [
     "Channel",
     "Variable",
     # From methods in alphabetic order
-    "from_andor",
     "from_BrunoldrRaman",
     "from_COLORS",
     "from_JASCO",
@@ -35,6 +34,7 @@ __all__ = [
     "from_PyCMDS",
     "from_ocean_optics",
     "from_shimadzu",
+    "from_Solis",
     "from_spcm",
     "from_Tensor27",
 ]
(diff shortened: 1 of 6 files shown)


WrightTools/data/_andor.py
```diff
@@ -0,0 +1,109 @@
...
+        Default is None.
+    parent : WrightTools.Collection (optional)
+        Collection to place new data object within. Default is None.
+    verbose : boolean (optional)
+        Toggle talkback. Default is True.
+
+    Returns
+    -------
+    data
+        New data object(s).
```
untzag: this function should always return one data object---clarify that here

WrightTools/data/_andor.py
```diff
@@ -0,0 +1,109 @@
...
+        axis0 = []
+        arr = []
+        attrs = {}
+        while True:
+            line = f.readline().strip()[:-1]
+            if len(line) == 0:
+                break
+            else:
+                line = line.split(',')
+                line = [eval(x) for x in line]
```
untzag: `eval` considered harmful...

do we need `eval`? I've just got to imagine that there is a better way to parse this info

at the very least, can we use [`ast.literal_eval`](https://docs.python.org/2/library/ast.html#ast.literal_eval) instead?

p770193:
@wright-group/wrighttools-maintainers ready for re-review.

WrightTools/data/__init__.py
```diff
@@ -26,6 +27,7 @@
     "Channel",
     "Variable",
     # From methods in alphabetic order
+    "from_andor",
```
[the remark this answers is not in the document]
untzag: this actually breaks import, so I cannot test how the docs work right now---next round :smile:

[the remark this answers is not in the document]
p770193: sorry about that...should be good now

WrightTools/data/_solis.py
```diff
@@ -0,0 +1,110 @@
...
+
+def from_Solis(filepath, name=None, parent=None, verbose=True):
+    """Create a data object from Andor Solis software (ascii exports).
+
+    Parameters
+    ----------
+    filepath : string, list of strings, or array of strings
+        Path to .txt file.
+    name : string (optional)
+        Name to give to the created data object. If None, filename is used.
```
ksunden: Does file metadata provide a name? if so, that is preferred to the file name itself

WrightTools/data/_solis.py
```diff
@@ -0,0 +1,110 @@
...
+    data.create_variable(name='ypos', values=np.arange(arr.shape[1])[None, :], units=None)
+    data.transform(data.variables[0].natural_name, 'ypos')
+
+    for key, val in attrs.items():
+        data.attrs[key] = val
+
+    # finish
+    if verbose:
+        print("data created at {0}".format(data.fullpath))
+        print("  size: {0}".format(data.size))
```
ksunden: Is there something more informative to print? I feel like axes, shape are better than size (though size and shape will typically be the same always, at least for our current hardware, correct?)

WrightTools/data/_solis.py
untzag: I took a look, and doesn't seem like there is a name in the footer

guess it makes sense, as I don't imagine that the software requires users to specify a name at acquisition time

WrightTools/data/_solis.py
untzag: I agree

PR #662, opened 2018-07-11
remove apparently unnecessary hdf package installs

ksunden:


ksunden:
Failure was clip #553 rerunning

PR #664, opened 2018-07-12
change index natural name

p770193:
Solis variables that track array coordinates now have names `xindex` and `yindex`, instead of `ypos`  and `xpos`.  New names are more concise and saves the namespace of position for variables with spatial units.


p770193:
tests are updated

PR #665, opened 2018-07-12
fluence calculations (resolves #625)

### LINKED ISSUE #625: Fluence calculation
I oftentimes find myself needing to calculate a fluence from a power reading. I then go through ten minutes of unit conversions to do the calculation. I think we should make a function in `kit._calculate.py` that does this calculation.

For instance:
```
def fluence(power_mW, reprate_Hz, photon_energy_eV, beam_radius_mm):
      
     return uj/cm2, photons/cm2
```

Any thoughts?

darienmorrow:
~~I realize my test is lame~~


ksunden:
we may wish to look at `unyt` to simplify the computation here, perhaps.

I'm considering replacing the  custom code in the `units` package with that library, which has some potential positive side effects

ksunden:
Also, the pep8 warning still mentioned above are all in the docstring (which is why `black` didn't catch it)

I believe `autopep8` will catch that if you don't feel like tracking those down manually

ksunden:
https://github.com/ambv/black/issues/318 
(`black` will NOT handle this case, as there exist cases [perhaps poorly reasoned, but still] where doing so will affect correctness of the program, and `black` commits to not making any such changes)

darienmorrow:
@ksunden The reason I am not checking the value is because that requires me to be 100% sure about the correct value of the calculation. I am waiting for someone to check my math before I do that.

darienmorrow:
@ksunden The reason I do not check for the value in my test is because checking for the value requires me to be absolutely sure about the result of my calculation. I was hoping for someone to check my math before I specify values.

darienmorrow:
@ksunden I now check the value in the test.

ksunden:
Im holding off rereviewing until the math itself is approved, I would like to check thebither two outputs as well in the test, please.

tests/kit/fluence.py
```diff
@@ -14,4 +14,7 @@
 
 def test_0():
     out = wt.kit.fluence(1, 2, .1, 1000, 1, "eV", "cm", "ps_t")
-    assert np.isclose(31.8309, out[0], rtol=1e-3)
+    checks = (31.83098, 99336493460095.2, 0.03183098)
+    assert np.isclose(checks[0], out[0], rtol=1e-3)
+    assert np.isclose(checks[1], out[1], rtol=1e-3)
+    assert np.isclose(checks[2], out[2], rtol=1e-3)
```
ksunden: This is fine, but for future reference, `np.allclose` exists.

PR #669, opened 2018-07-18
kit.svd

darienmorrow:
- [x] add svd method to kit
- [x] test(s)
- [x] slice functionality @ksunden

### DIFF
--- a/WrightTools/kit/_array.py
+++ b/WrightTools/kit/_array.py
@@ -216,7 +216,7 @@ def remove_nans_1D(*args):
     vals = np.isnan(args[0])
     for a in args:
         vals |= np.isnan(a)
-    return tuple(np.array(a)[vals == False] for a in args)
+    return tuple(np.array(a)[vals is False] for a in args)
 
 
 def share_nans(*arrs):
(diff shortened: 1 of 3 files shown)


untzag:
[E203 is not actually PEP8 compliant](https://black.readthedocs.io/en/stable/the_black_code_style.html#slices)

WrightTools/kit/_array.py
```diff
@@ -215,7 +216,7 @@ def remove_nans_1D(*args):
     vals = np.isnan(args[0])
     for a in args:
         vals |= np.isnan(a)
-    return tuple(np.array(a)[vals == False] for a in args)
```
ksunden: this actually DOES need to be ==, should add a comment.... See #534

ksunden:
Approved pending completion of tests.

I do wish for a comment from someone else as to whether the `vals == False` should be refactored to `~vals`, however. That would allow for removal of the `noqa` and comment explaining it's existance.

untzag:
@ksunden sure, let's refactor---seems like a better choice

Issue #588, opened 2018-04-27
Chop uses both "expressions" and "name" to identifiy axes


ksunden:
https://github.com/wright-group/WrightTools/blob/490a4a3d6fb6f016e7033d661b553b72c2d86fcb/WrightTools/data/_data.py#L306-L319

This is a problem when expressions are used e.g. `w1=w2` where the "name" becomes `w1__e__w2` as these two are different, one of these checks *will* fail

ksunden:
in particular, lines 306 and 318 are of interest here

ksunden:
workaround (at least for the equal case): transform before performing the operation so that the name and expression match (removing the `=`)

PR #670, opened 2018-07-19
interact2D

### LINKED ISSUE #661: interactive 2D plots
- [x] Interactive side plots, where mouse clicks on 2D subplot reveal the slices on the side plots
- [x] With 3D+ data, interactive sliders are used to select the 2D slice to be shown.

### LINKED ISSUE #679: pcolor:  bare arrays do not use pcolor_helper
In the artists.Axis subclass, `pcolor_helper` is called for bare arrays, but it does not pass these arguments to `pcolor`.  
Need to explicitly reassign args:  `args[:3] = X, Y, Z`.

Also need to transpose X and Y to match shape of Z

### LINKED ISSUE #680: Chop gives different slice if data object is transformed
```python
import numpy as np
from matplotlib import pyplot as plt
import WrightTools as wt

x = np.arange(6)
y = x[::2].copy()
z = np.arange(x.size * y.size).reshape(x.size, y.size).astype("float")
z[:, y < 2] *= 0
data = wt.data.Data(name="data")
data.create_channel("signal", values=z, signed=False)
data.create_variable("x", values=x[:, None], units="wn")
data.create_variable("y", values=y[None, :], units="wn")
data.x.label='x'
data.y.label='y'
data.transform("x", "y")

data.print_tree()

d = data.chop('x', at={'y':(2, 'wn')})
d.print_tree()

data.transform("y", "x")
d = data.chop('x', at={'y':(2, 'wn')})
d.print_tree()
```
Gives the following output (to stdout, stderr has unrelated warnings):
```
data (/tmp/vnt3uxjk.wt5)
├── axes
│   ├── 0: x (wn) (6, 1)
│   └── 1: y (wn) (1, 3)
├── variables
│   ├── 0: x (wn) (6, 1)
│   └── 1: y (wn) (1, 3)
└── channels
    └── 0: signal (6, 3)
chopped data into 1 piece(s) in ['x', 'y']
chop (/tmp/w2kyp0qh.wt5)
└── 0: chop000 (6,)
    ├── axes: x (wn)
    └── channels: signal
chopped data into 1 piece(s) in ['x', 'y']
chop (/tmp/64c3gwp6.wt5)
└── 0: chop000 (3,)
    ├── axes: x (wn)
    └── channels: signal
```

Note that the two calls to chop are identical, the only difference is the order of axes in the data object itself.
The second call is incorrect, as a slice with axis `x` should have shape `(6,)`, as it does in the first call.

p770193:
resolves #661, resolves #679, resolves #680

- [x] basic layout of figure
- [x] signed projections:  show the envelope (negative and positive projections) rather than the total sum
- [x] axis decorators (axes labels, axes positions label for sliders, colorbar)
- [x] make global/local button actually work


WrightTools/artists/_quick.py
```diff
@@ -148,7 +148,7 @@ def quick2D(
     autosave=False,
     save_directory=None,
     fname=None,
-    verbose=True
+    verbose=True,
```
ksunden: Extra comma is valid in py3.6, but not 3.5

p770193: thanks for pointing this out--comma is now removed

untzag:
@p770193 as far as I can tell, there aren't any good "drop down" menus for selecting channels

for now, let's just ignore that feature?

untzag:
@p770193 I kind of don't like the `teal` and `coral`---could we maybe just make the cursor lines dark grey?

p770193:
@ksunden can you help me understand why the Travis check failed?  It seems like it had to do with tests unrelated to my changes.

ksunden:
Failure was clip, see #553, rerunning

ksunden:
We had another clip #553 error. Just trying to document how often they occur

darienmorrow:
`interact2D` does not work if `xaxis`, `yaxis`, or `channel` are strings. Something like `channel_index = wt_kit.get_index(data.channel_names, channel)` should probably used.

darienmorrow:
Do we desire to put ticks on the slider bars?

darienmorrow:
Can we handle infinities in channels?

darienmorrow:
Should we use ax.hline and ax.vline ?

darienmorrow:
Have title with data object name?

darienmorrow:
As far as my tests go, we still don't have the correct access to different channels and axes by passing strings. For instance,
https://github.com/wright-group/WrightTools/blob/c3cf73daff5db2b1eb6a3ab70a54e748bab54baa/WrightTools/artists/_interactive.py#L44
does not work as intended because `channel` is a string, not a data object. 

I vote that we use https://github.com/wright-group/WrightTools/blob/f22920579f45632b4123661d9832ff0cc1b614c4/WrightTools/kit/_list.py#L74 within the `get_channel` and `get_axes` methods.

p770193:
@darienmorrow that was due to a typo, sorry.  It should work as you want it now.  I have not used `get_index` because it does not do what I want.

p770193:
`interact2D` uses discrete slider steps with the `matplotlib.widgets.Slider` object.  This discrete stepping was only made available in [matplotlib version 2.2](https://matplotlib.org/users/whats_new.html#slider-ui-widget-can-snap-to-discrete-values).  @ksunden is it appropriate for WrightTools to require mpl 2.2 now (currently requires 2.0)?

darienmorrow:
@ksunden, As requested by @p770193, I have messed around with this PR. Everything now seems to work as expected. I even found a matplotlib version issue by playing around with it on a different machine (see above comment). I think the PR is ready for your review!

WrightTools/artists/_interact.py
```diff
@@ -0,0 +1,381 @@
...
+                / + "narrow.".format(aspect)
+            )
+            aspect = np.clip(aspect, 1 / 3., 3.)
+            print("using aspect {0} instead".format(aspect))
+    else:
+        aspect = 1
+    return aspect
+
+
+def interact2D(data, xaxis=0, yaxis=1, channel=0, local=False, verbose=True):
```
ksunden: Do we wish to allow `**kwargs` and pass those to the lower level plots?
Not sure we do, but want to consider it.

ksunden: I'm thinking things like colormaps and such (though the Qt viewer may provide some access as well)

WrightTools/artists/_interact.py
```diff
@@ -0,0 +1,381 @@
...
+    """
+    # unpack
+    channel = get_channel(data, channel)
+    xaxis, yaxis = get_axes(data, [xaxis, yaxis])
+    cmap = get_colormap(channel)
+    current_state = Bunch()
+    # create figure
+    nsliders = data.ndim - 2
+    if nsliders < 0:
+        print("note enough dimensions")
```
ksunden: raise exception, rather than print and return `wt.exceptions.DimensionalityError` exists for this exact purpose

WrightTools/artists/_interact.py
```diff
@@ -0,0 +1,381 @@
...
+        return
+    # TODO: implement aspect; doesn't work currently because of our incorporation of colorbar
```
ksunden: Resolved?

WrightTools/artists/_interact.py
```diff
@@ -0,0 +1,381 @@
...
+    fig, gs = create_figure(width="single", nrows=7 + nsliders, cols=[1, 1, 1, 1, 1, "cbar"])
```
ksunden: why 7 rows? seems like this would be better suited to the `aspect` kwarg of `create_figure`. 
Same question applies to the number of columns.

WrightTools/artists/_interact.py
```diff
@@ -0,0 +1,381 @@
...
+    # create axes
+    ax0 = plt.subplot(gs[1:6, 0:5])
+    ax0.patch.set_facecolor("w")
+    ax0.grid(b=True)
+    cax = plt.subplot(gs[1:6, -1])
+    sp_x = add_sideplot(ax0, "x", pad=0.1)
+    sp_y = add_sideplot(ax0, "y", pad=0.1)
+    ax_local = plt.subplot(gs[0, 0], aspect='equal', frameon=False)  # facecolor='#EEEEEE')
```
ksunden: Should this comment be deleted?

WrightTools/artists/_interact.py
```diff
@@ -0,0 +1,381 @@
...
+                  transform=ax_title.transAxes)
+    ax_title.set_axis_off()
+    # NOTE: there are more axes here for more buttons / widgets in future plans
+    # create lines
+    x_color = "#00BFBF"  # cyan with saturation increased
+    y_color = "coral"
+    line_sp_x = sp_x.plot([None], [None], visible=False, color=x_color)[0]
+    line_sp_y = sp_y.plot([None], [None], visible=False, color=y_color)[0]
+    crosshair_hline = ax0.plot([None], [None], visible=False, color=x_color)[0]
+    crosshair_vline = ax0.plot([None], [None], visible=False, color=y_color)[0]
```
ksunden: Any reason these are `plot` and not `ax[hv]line`?

WrightTools/artists/_interact.py
```diff
@@ -0,0 +1,381 @@
...
+                    colors='k', linestyle=':', alpha=0.5
+            )
+    # initial xyz start are from zero indices of additional axes
+    slices = get_slices(sliders, data.axes, verbose=verbose)
+    zi = channel[slices]
+    zi = zi.squeeze()
+    if wt_kit.get_index(data.axes, xaxis) < wt_kit.get_index(data.axes, yaxis):
+        zi = zi.T.copy()
+    current_state.slices = slices
+    # TODO: should we use pcolormesh or pcolor?
```
ksunden: I would lean pcolor, as it is more general, and I'm not sure we can assume even spacing.

ksunden: pcolormesh is faster, though, and speed is important with interactivity, so willing to be convinced otherwise, especially if it impacts user experience.

WrightTools/artists/_interact.py
```diff
@@ -0,0 +1,381 @@
...
+        slices = get_slices(sliders, data.axes, verbose=verbose)
+        if verbose:
+            print(index)
+        if radio.value_selected.strip() == "global":
+            current_state.local = False
+        if radio.value_selected.strip() == "local":
+            current_state.local = True
+        arr = channel[slices].squeeze()
+        clim = get_clim(channel, current_state.local, arr)
+        print(clim)
```
ksunden: extraneous print

ksunden: (or needs if verbose: and a string saying what it is)

WrightTools/artists/_interact.py
```diff
@@ -0,0 +1,381 @@
...
+        y_temp = np.abs(yaxis.points - y0)
+        y_index = np.argmin(y_temp)
+        side_plot = arr[y_index].copy()
+        side_plot = norm(side_plot, channel.signed)
+        line_sp_x.set_data(xaxis.points, side_plot)
+
+    def update_local(index):
+        slices = get_slices(sliders, data.axes, verbose=verbose)
+        if verbose:
+            print(index)
```
ksunden: Consider adding a string to explain what is being printed

WrightTools/artists/_interact.py
```diff
@@ -0,0 +1,381 @@
...
+        if radio.value_selected.strip() == "global":
+            current_state.local = False
+        if radio.value_selected.strip() == "local":
+            current_state.local = True
```
ksunden: four lines can be condensed to `current_state.local = radio.value_selected.strip() == "local"`
is strip necesary? Seems like you are comparing to known strings, not user input.

WrightTools/artists/_interact.py
```diff
@@ -0,0 +1,381 @@
...
+
+    def update(info):
+        # is info a value?  then we have a slider
+        # is info an object with xydata?  then we have an event
+        slices = get_slices(sliders, data.axes, verbose=verbose)
+        if slices != current_state.slices:  # a Slider moved; need to update all plot objects
+            arr = channel[slices].squeeze()
+            if wt_kit.get_index(data.axes, xaxis) < wt_kit.get_index(data.axes, yaxis):
+                arr = arr.T.copy()
+            # TODO: why am I stripping off array information?
```
ksunden: resolved?

WrightTools/artists/_interact.py
```diff
@@ -0,0 +1,381 @@
...
+        clim = get_clim(channel, current_state.local, arr)
+        print(clim)
+        ticklabels = gen_ticklabels(np.linspace(*clim, 11))
+        colorbar.set_ticklabels(ticklabels)
+        obj2D.set_clim(*clim)
+        fig.canvas.draw_idle()
+
+    def update(info):
+        # is info a value?  then we have a slider
+        # is info an object with xydata?  then we have an event
```
ksunden: would prefer if these comments were not phrased as questions, as I initially read them as meaning the programmer was unclear of what `info` is, when in fact the opposite is true. 
That said, these comments seem disconnected from the nearby code. Should they be moved or deleted?

WrightTools/artists/_interact.py
```diff
@@ -0,0 +1,381 @@
...
+        slices = get_slices(sliders, data.axes, verbose=verbose)
+        if slices != current_state.slices:  # a Slider moved; need to update all plot objects
+            arr = channel[slices].squeeze()
+            if wt_kit.get_index(data.axes, xaxis) < wt_kit.get_index(data.axes, yaxis):
+                arr = arr.T.copy()
+            # TODO: why am I stripping off array information?
+            # cf. https://stackoverflow.com/questions/29009743
+            obj2D.set_array(arr[:-1, :-1].ravel())
```
ksunden: if pcolor is used rather than pcolormesh, may wish to simplify this line

WrightTools/artists/_interact.py
```diff
@@ -0,0 +1,381 @@
...
+            sp_y.collections.clear()
+            draw_sideplot_projections(arr)
+            current_state.slices = slices
+            if line_sp_x.get_visible() and line_sp_y.get_visible():
+                update_sideplot_slices()
+        elif info.inaxes == ax0:  # crosshairs
+            x0 = info.xdata
+            y0 = info.ydata
+            if x0 is None or y0 is None:
+                print(info, info.xydata)
```
ksunden: any information wished to be printed upon exception raising should be passed to that exception.

WrightTools/artists/_interact.py
```diff
@@ -0,0 +1,381 @@
...
+                raise AttributeError
```
ksunden: please use a WrightTools derived exception

WrightTools/artists/_quick.py
```diff
@@ -10,7 +10,7 @@
 
 import matplotlib.pyplot as plt
 
-from ._helpers import create_figure, plot_colorbar, savefig
+from ._helpers import create_figure, plot_colorbar, savefig, add_sideplot
```
ksunden: Is this now an unused import?

WrightTools/artists/_interact.py
```diff
@@ -0,0 +1,381 @@
...
+
+def set_aspect(xaxis, yaxis):
+    if xaxis.units == yaxis.units:
+        xr = xaxis.max() - xaxis.min()
+        yr = yaxis.max() - yaxis.min()
+        aspect = np.abs(yr / xr)
+        if 3 < aspect or aspect < 1 / 3.:
+            raise Warning(
+                "units agree, but aspect {0} required for equal spacing is too"
+                / + "narrow.".format(aspect)
```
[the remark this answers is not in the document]
p770193: while once used in this branch, `set_aspect` is an unused function currently.  I'll just remove it and we can ignore comments involving it.

WrightTools/artists/_interact.py
p770193: Perhaps we can consider this an enhancement to be added at a later date?

WrightTools/artists/_interact.py
p770193: no

ksunden: The TODO can stay as a comment, but should have an associated issue (I just don't want to loose track of things we intend to do)

ksunden: I would consider adding a date/initials to the TODO, just so that thought is attributable (git serves this purpose as well, but visibility is useful/the line could be edited for something like whitespace or refactoring which would obfuscate the git history)

WrightTools/artists/_interact.py
p770193: We tried to use aspect but had issues when incorporating the colorbar (cf. 53d1f1faa684c49a556e03938c41e4157b692f1a).  I think `aspect` failed because `create_figure` fixes the aspect for all elements in a row.  When gridspec has a regular axis and a colorbar axis it doesn't work well. @untzag can you confirm?

ksunden: Aspect works with colorbars, as far as I know, I use it in publication figures I know... 
Maybe sliders are the culprit?
If aspect acts as documented, which I believe, it affects the first column.

WrightTools/artists/_interact.py
p770193: I don't think so...I think I have to borrow `pcolor_helper` to get this to work correctly.  Good catch

WrightTools/artists/_interact.py
p770193: Nothing is tested with pcolor vs. pcolormesh yet, but I will say speed is on my mind, especially with dense 2D datasets such as we will have with ccd images.  We should investigate further, but I vote we leave it for now

WrightTools/artists/_base.py
```diff
@@ -406,6 +406,81 @@ def pcolor(self, *args, **kwargs):
...
+            xlabel=kwargs.pop("xlabel", None),
+            ylabel=kwargs.pop("ylabel", None),
+            data=data,
+            channel_index=channel_index,
+        )
+        # decoration
+        self.set_facecolor([0.75] * 3)
+        # call parent
+        return super().pcolormesh(*args, **kwargs)
+
```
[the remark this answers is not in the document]
p770193: I will note that I potentially need to steal some code from `pcolor_helper` to update `PolyCollection` object I am working with.  We may want to discuss a new division of functions.

WrightTools/artists/_interact.py
ksunden: I would say either `TypeError`

`AttributeError` tends to specifically mean the call to `__getattr__` (which is called by doing `foo.bar`) fails. That is not the case here, as the attribute of info is defined (or at least does not raise an exception)
(Python docs)[https://docs.python.org/3.7/library/exceptions.html#AttributeError]

TypeError is appropriate as x0/y0 have NoneType, not the type expected (I beleive numeric, maybe float specifically). We define in exceptions.py a TypeError which inherits from both our own base exception and the standard library TypeError, so it can be caught by a calling function by either.

ValueError is a similar catchall, but is defined as being the correct type, which I do not think applies here.

I am not opposed to adding AttributeError to exceptions.py. Mostly the Idea is to have a base exception class from which all exceptions thrown by WrightTools itself can be caught. We did not define all of the standard lib exception, as we do not anticipate needing most of them.

Admittedly for the current scope of where interact2D is anticipated to be used, this kind of boxing of exceptions seems slightly superfluous. However, I'd like to have these systems in place and make sure it can be called in some larger application as well (perhaps as a pane in a larger GUI app)

WrightTools/artists/_interact.py
ksunden: I can agree to this

WrightTools/artists/_interact.py
```diff
@@ -331,8 +326,7 @@ def update(info):
             x0 = info.xdata
             y0 = info.ydata
             if x0 is None or y0 is None:
-                print(info, info.xydata)
-                raise AttributeError
+                raise AttributeError(info)
```
ksunden: Commented this on a now outdated version, repeating for visibility:

I would say  TypeError

AttributeError tends to specifically mean the call to __getattr__ (which is called by doing foo.bar) fails. That is not the case here, as the attribute of info is defined (or at least does not raise an exception)
(Python docs)[https://docs.python.org/3.7/library/exceptions.html#AttributeError]

TypeError is appropriate as x0/y0 have NoneType, not the type expected (I beleive numeric, maybe float specifically). We define in exceptions.py a TypeError which inherits from both our own base exception and the standard library TypeError, so it can be caught by a calling function by either.

ValueError is a similar catchall, but is defined as being the correct type, which I do not think applies here.

I am not opposed to adding AttributeError to exceptions.py. Mostly the Idea is to have a base exception class from which all exceptions thrown by WrightTools itself can be caught. We did not define all of the standard lib exception, as we do not anticipate needing most of them.

Admittedly for the current scope of where interact2D is anticipated to be used, this kind of boxing of exceptions seems slightly superfluous. However, I'd like to have these systems in place and make sure it can be called in some larger application as well (perhaps as a pane in a larger GUI app)

ksunden:
@p770193 Thanks for your careful attention to all my comments. This will be great to have.

WrightTools/artists/_base.py
```diff
@@ -332,50 +332,28 @@ def legend(self, *args, **kwargs):
...
-        **kwargs
-            matplotlib.axes.Axes.pcolor__ optional keyword arguments.
-
-            __ https://matplotlib.org/api/_as_gen/matplotlib.axes.Axes.pcolor.html
-
-        Returns
-        -------
-        matplotlib.collections.PolyCollection
-        """
+    def _parse_plot_args(self, *args, **kwargs):
```
p770193: This merges the duplicitous code between pcolor and pcolormesh.  It seems like all our other plotting commands could be sent here as well to further simplify, but perhaps I make an issue for that and address those concerns with another PR.

WrightTools/artists/_base.py
```diff
@@ -332,50 +332,28 @@ def legend(self, *args, **kwargs):
...
-        # unpack data object, if given
         if isinstance(args[0], Data):
             data = args.pop(0)
-            if not data.ndim == 2:
+            if plot_type in ['pcolor', 'pcolormesh', 'contour', 'contourf']:
+                ndim = 2
+            elif plot_type in ['plot']:
+                ndim = 1
+            if not data.ndim == ndim:
                 raise wt_exceptions.DimensionalityError(2, data.ndim)
```
ksunden: ndim rather than 2

WrightTools/artists/_base.py
```diff
@@ -332,50 +332,28 @@ def legend(self, *args, **kwargs):
...
         args = list(args)  # offer pop, append etc
         channel = kwargs.pop("channel", 0)
         dynamic_range = kwargs.pop("dynamic_range", False)
-        # unpack data object, if given
         if isinstance(args[0], Data):
             data = args.pop(0)
-            if not data.ndim == 2:
+            if plot_type in ['pcolor', 'pcolormesh', 'contour', 'contourf']:
+                ndim = 2
+            elif plot_type in ['plot']:
```
ksunden: Is this intended to be used for `plot`?
If so, there are some parts of the method (e.g. lines 350 - 360 or so) which assume 2D axes

ksunden: @p770193 (just bumping because it had gotten collapsed)

WrightTools/artists/_base.py
ksunden: I think I agree that it should hold off to send others here, but it's a good start. Thanks

ksunden:
Okay, I think most of my comments have been addressed.
Just a few last clean up things:
I'd like to run these files through `black` to ensure consistent formatting (I can do this, but want to be sure the code isn't changing under me when I do)
We should update the `setup.py` requirements for `matplotlib>=2.2` rather than `2.0` as we are now reliant on that kwarg which was introduced later.

p770193:
@ksunden please note my newest changes, and (if appropriate) proceed with `black`, `setup.py`

p770193:
@ksunden, I think we need some user testing before merging.  I just found a bug introduced in my recent commits

WrightTools/artists/_interact.py
```diff
@@ -201,7 +203,8 @@ def interact2D(data, xaxis=0, yaxis=1, channel=0, local=False, verbose=True):
     current_state.zi = zi
     clim = get_clim(channel, current_state)
     xi, yi, zi = pcolor_helper(xaxis.points[:, None], yaxis.points[None, :], zi)
-    obj2D = ax0.pcolormesh(xi, yi, zi, cmap=cmap, vmin=clim[0], vmax=clim[1])
+    plt.sca(ax0)
+    obj2D = plt.pcolormesh(xi, yi, zi, cmap=cmap, vmin=clim[0], vmax=clim[1])
```
ksunden: Why is this necessary? ax.pcolormesh should be able to handle bare arrays as well as data objects.
If it does not, it should be fixed to do so as that is intended behavior

ksunden: @p770193 (bumping because this got collapsed in `black` commit)

ksunden: I also hesitate to have unexpected side effects within this function (the axes pointed to by plt could be changed for a user)

WrightTools/artists/_helpers.py
```diff
@@ -487,7 +487,7 @@ def get_scaled_bounds(ax, position, *, distance=0.1, factor=200):
     return [h_scaled, v_scaled], [va, ha]
 
 
-def pcolor_helper(xi, yi, zi):
+def pcolor_helper(xi, yi):
```
ksunden: for backwards compatibility, we may wish to include `zi=None`, with an entry in the docstring stating it does nothing.
We could also invoke a DepricationWarning if it is not `None`

ksunden: pcolor_helper _is_ part of our public API, removing the third positional argument is a breaking change to old code.

WrightTools/artists/_helpers.py
```diff
@@ -543,7 +548,13 @@ def full(arr):
             ll = orig[idx[0] + 0, idx[1] + 0]
             lr = orig[idx[0] + 0, idx[1] + 1]
             out[idx] = np.mean([ul, ur, ll, lr])
-    return X, Y
+    if zi is not None:
+        warnings.warn(
+            "zi argument is not used in pcolor_helper and is not required", DeprecationWarning
```
ksunden: Can we use wt_exceptions.VisibleDeprecationWarning instead? By default DeprecationWarnings are silent

ksunden:
Test failure is choice of name over expression in commit which references #588

darienmorrow:
https://github.com/wright-group/WrightTools/blob/59da5f88eec609343ad8551743431ad875b19ff4/WrightTools/artists/_interact.py#L134-L145
Copying and removing of variables and channels should all have `verbose=False` added as a kwarg.

ksunden:
This pull request **introduces 4 alerts** and **fixes 6** when merging be7d4ec6c4dbe42ece5af74a96019240d0e15f03 into f22920579f45632b4123661d9832ff0cc1b614c4 - [view on LGTM.com](https://lgtm.com/projects/g/wright-group/WrightTools/rev/pr-decca392abbf850a3d72702682676c55b92eb677)

**new alerts:**

* 3 for Unused local variable
* 1 for \_\_eq\_\_ not overridden when adding attributes

**fixed alerts:**

* 6 for Unused local variable

---

*Comment posted by [LGTM.com](https://lgtm.com)*

ksunden:
This pull request **introduces 4 alerts** and **fixes 6** when merging f6f7b373abe6cd5651a9420f85efa8c2c46cb6ac into f22920579f45632b4123661d9832ff0cc1b614c4 - [view on LGTM.com](https://lgtm.com/projects/g/wright-group/WrightTools/rev/pr-997b5c094e6a442993b2313f87e0bbd907317728)

**new alerts:**

* 3 for Unused local variable
* 1 for \_\_eq\_\_ not overridden when adding attributes

**fixed alerts:**

* 6 for Unused local variable

---

*Comment posted by [LGTM.com](https://lgtm.com)*

ksunden:
This pull request **introduces 4 alerts** and **fixes 6** when merging 905874182e7532e8e84cafca3148a2bb81c8cb41 into f22920579f45632b4123661d9832ff0cc1b614c4 - [view on LGTM.com](https://lgtm.com/projects/g/wright-group/WrightTools/rev/pr-98a0f44345999847bbc8ffd068d34b466d0fd4c6)

**new alerts:**

* 3 for Unused local variable
* 1 for \_\_eq\_\_ not overridden when adding attributes

**fixed alerts:**

* 6 for Unused local variable

---

*Comment posted by [LGTM.com](https://lgtm.com)*

ksunden:
This pull request **introduces 4 alerts** and **fixes 6** when merging 086b868423b8ebe7c3351d68f4b83eb33c661796 into f22920579f45632b4123661d9832ff0cc1b614c4 - [view on LGTM.com](https://lgtm.com/projects/g/wright-group/WrightTools/rev/pr-d19d682f1dbecc063f93bd76910ed5e797d745a2)

**new alerts:**

* 3 for Unused local variable
* 1 for \_\_eq\_\_ not overridden when adding attributes

**fixed alerts:**

* 6 for Unused local variable

---

*Comment posted by [LGTM.com](https://lgtm.com)*

darienmorrow:
@p770193 @ksunden Do we have a checklist for what is still needed on this PR?

ksunden:
@darienmorrow I think it is now ready as far as I'm concerned.

We may revisit to optimize at some point, but it is usable and works for all the datasets we've thrown at it, as far as I know. That is my criterion for merging.

Pending passing tests (As I did just modify the tests, though they did work on my machine)
Perhaps someone should look over my changes.

PR #698, opened 2018-08-04
Implementation of split in WT3

### LINKED ISSUE #474: recover split

ksunden:
Closes #474 

Differences from WrightTools 2 era split:

* No longer has behavior where it "snaps" to the nearest point. All splits are now the evaluation of the expression given in the units of that expression  (if given as an axis index, the axis units are give, otherwise defaults to the units of the first variable)
* Split include the lower bound, but exclude the upper bound (no `direction` kwarg)
* As it returns a collection, no longer returns `None` for empty data object, but does not populate any variables or channels if that is the case (equivalent to `wt.data.Data(name)`)
* No longer lowers the rank (ndim) of the data object if a single slice is taken in any dimension (This one may change if we wish for it to, but current implementation does not)
* Can now split on arbitrary expressions of variables, with arbitrary shapes.
* Outputs are now ordered in ascending order of split positions in the expression given. (Rather than array order, which is simply not meaningful as arbitrary splits are allowed) Can discuss whether we want to flip the order if a list is given in descending order. My current inclination is to say sorted order is most clear and is deterministic which is my first goal.

What's left to do:

- [x] verbose output
- [x] Documentation (docstring mostly there, but would like to add to the human generated pages, gallery, etc)
- [x] Additional tests of new behaviors
- [x] Decision on lowering rank of resultant data objects
- [x] Decision on descending list
- [x] Decision on whether or not we wish to add a "tolerance" kwarg, and what it's exact behavior would be
- [x] Either work around the `keepdims` kwarg or make numpy 1.15 the minimum dependency. 
- [x] figure out odd behavior of one particular split (mostly a note for myself)
- [x] Split up mask reduction helper method into two, one to reduce the mask, one to enforce a shape.


WrightTools/data/_data.py
```diff
@@ -1348,92 +1343,95 @@ def split(self, axis, positions, units="same", direction="below", parent=None, v
...
+            "s -> (s0,s1), (s1,s2), (s2, s3), ..."
+            a, b = itertools.tee(iterable)
+            next(b, None)
+            return zip(a, b)
+
+        def mask_reduce(mask, shape):
+            mask = mask.copy()
+            for i in range(len(mask.shape)):
+                a = mask.copy()
+                offset = 0
```
ksunden: This can go away, is a remnant of a previous iteration of this function.

ksunden:
Please do note that this requires numpy 1.15

ksunden:
Note, that that odd behavior mentioned above appears to be an issue of plotting, not of the split

ksunden:
Odd behavior was because that particular dataset had an all NaN chunk, and the reduction after chunkwise was not accounting for NaNs

ksunden:
My personal leanings on the above decision tasks:

Better for users to do the reduction. (Not too hard to implement, but not completely trivial)

Do not flip for descending list

tolerance kwarg is easily added at a later date, would like to demonstrate need

Pin numpy to >= 1.15.0. Numpy is pretty darn good about not regressing, so don't think there is a high likelihood of conflict with maximum dependency of another library. While the think I'm using could be worked around, I think the built in method of doing it will be more efficient.


I'm going to check off the line items that require no further code changes. If other users have issues with these decisions, please comment here.

ksunden:
One question I have for the reviewer: Currently verbose output prints the shape of the particular expression that was split against, not of the resultant data object. Should this be changed?

Issue #678, opened 2018-07-27
from_spcm:  import metadata


p770193:
spcm files may include many details about the experiment in the header.  Search for these details and import them when available.  

Also convert signal channel to units of counts/sec, in the same style as from_Solis

ksunden:
What units is the signal channel in already?

I'd recommend adding a channel rather than simple replacement, just for tracability's sake

p770193:
currently units of signal are counts (not normalized by collection time).  I get the tracability, but it feels silly to me to keep both time-normalized and not time-normalized channels, when you can convert from one to the other with metadata...I guess whether or not to include it depends on exactly how much this import method has been used to this point.

ksunden:
To me, the `from_<x>` methods should be as directly related to the file it represents. As minimal as possible modifications/number crunching.

p770193:
I can agree with having two channels, time normalized and not (although this does seem a little silly since the relation between the two is trivial).  Or can time-normalization be a kwarg in the from method?

p770193:
Ended up not normalizing by time--this step is left to the user.  Among the reasons for leaving it to the user:  the metadata gives the set collection time but not the true collection time (the two differ if a scan is interrupted, for instance).

PR #702, opened 2018-08-07
From spcm

p770193:
Resolves issue #678 .


WrightTools/data/_spcm.py
```diff
@@ -23,8 +23,9 @@
 # --- from function -------------------------------------------------------------------------------
 
 
-def from_spcm(filepath, name=None, *, delimiter=",", format=None, parent=None, verbose=True):
-    """Create a data object from Becker & Hickl `spcm`__ software.
+def from_spcm(filepath, name=None, *, delimiter=",", parent=None, verbose=True):
+    """Create a ``Data`` object from a Becker & Hickl `spcm`__ file (ASCII-exported, ``.asc``).
+    If provided, setup parameters are stored in the ``attrs`` dictionary of the ``Data`` object.
```
ksunden: Need a new line (for docstrings to be properly rendered on ReadTheDocs

WrightTools/data/_spcm.py
```diff
@@ -37,9 +38,6 @@ def from_spcm(filepath, name=None, *, delimiter=",", format=None, parent=None, v
         Default is None.
     delimiter : string (optional)
         The string used to separate values. Default is ','.
-    format : {'ascii'} (optional)
-        Force file to be interpreted as a specific format. Default is None
-        (autorecognized).
```
ksunden: Was format used?, I didn't see it, but just checking to be sure

WrightTools/data/_spcm.py
```diff
@@ -74,19 +68,61 @@ def from_spcm(filepath, name=None, *, delimiter=",", format=None, parent=None, v
...
+                    splitted = line.split()
+                    value = splitted[-1][1:-1].split(",")
+                    key = " ".join(splitted[:-1])
+                    headers[key] = value
+            line = f.readline().strip()
+            if "END" in line:
+                header_lines += 1
+                break
+    # initialize data object
+    kwargs = {"name": name, "kind": "spcm", "source": filepath, **headers}
```
ksunden: If we have access to the time of collection, it is intended behavior that the data object "created" date use that time. See `_pycmds.py` for an example of such. (Just a matter of passing in the kwarg/ensuring the date format is what we want) 

Here, it seems to be the combination of two header fields "Date" and "Time"

WrightTools/data/_spcm.py
p770193: format is not used

WrightTools/data/_spcm.py
```diff
@@ -74,27 +70,54 @@ def from_spcm(filepath, name=None, *, delimiter=",", format=None, parent=None, v
...
+                    headers[key] = value
+            line = f.readline().strip()
+            if "END" in line:
+                header_lines += 1
+                break
+    if "Date" in headers.keys() and "Time" in headers.keys():
+        # NOTE:  reports created in local time, no-way to calculate absolute time
+        created = " ".join([headers["Date"], headers["Time"]])
+        created = time.strptime(created, "%Y-%m-%d %H:%M:%S")  
+        created = timestamp.TimeStamp(time.mktime(created)).RFC3339
```
ksunden: You get the created time here, but looks like you never do anything with it should be added to the kwargs as the "special/reserved" attribute `"created"`

WrightTools/data/_spcm.py
darienmorrow: What timezone would the investment be in?

PR #717, opened 2018-08-15
Ensure proper dtype support in collapse

ksunden:
Also ensures that actions which would result in single numbers keep their dimensionality


ksunden:
The problem here is that operations on arrays with dtype int should convert to float, but we also wish to support complex....

ksunden:
should use https://docs.scipy.org/doc/numpy/reference/generated/numpy.result_type.html

ksunden:
min/max should preserve type regardless

PR #726, opened 2018-08-20
Proper source of the generic "Group"

### LINKED ISSUE #712: Group is not defined in collection
https://github.com/wright-group/WrightTools/blob/ca056aa600f341501a99d2ea4d11f7d74047bc26/WrightTools/_open.py#L48

Statement will cause an attribute error. Not tested currently

ksunden:
Closes #712

### DIFF
--- a/WrightTools/_open.py
+++ b/WrightTools/_open.py
@@ -10,6 +10,7 @@ import h5py
 
 from . import collection as wt_collection
 from . import data as wt_data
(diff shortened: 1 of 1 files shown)


ksunden:
found it by looking at the lines which are not covered by our test suite. Still not covered, but will write a test case when I have a use case for using the bare group in mind

Issue #684, opened 2018-08-02
interact2D:  additional documentation


p770193:
A general overview of the interact2D functionality should be made on ReadTheDocs.

PR #729, opened 2018-08-22
artist cleanup

### LINKED ISSUE #545: quick1D and quick2D features
The old `mpl1D` and `mpl2D` classes when plotting data taken from data object of higher dimensionality than the plot would have titles reflecting the slice through the non-plotted axis that was taken in the particular figure. The new `quick1D` and `quick2D` functions do not seem to have this feature. Is this intended?  

Is it also intended that quick2D doesn't have gridlines or the grey overline showing the non-scanned dimension if appropriate?

### LINKED ISSUE #667: Test artist methods
https://github.com/matplotlib/pytest-mpl

Pytest plugin to help, recommended by matplotlib core contributor (also the release maintainer of h5py)

darienmorrow:
closes #545 and closes #667

Todos :
- [x]  titles reflecting the slice through the non-plotted axis 
- [x] gridlines
- [x] grey overline showing the non-scanned dimension if appropriate
- [x] simple, minimally viable tests for `quick1D` and `quick2D`

Todos for @ksunden:
- [x] make links in docs
- [x] make crosshairs work
- [x] rcparams 
- [x] `plt.sca(ax)` ? 

Todos for @untzag or @p770193:
- [ ] read through the new `artists` docs


darienmorrow:
@ksunden There isn't an explicit test for `quick1D` and `quick2D` working in the minimal sense. Do we ever use them in the test suite?

untzag:
> titles reflecting the slice through the non-plotted axis

how would / could this work? it seems pretty hard to me

ksunden:
quick1d/2d are used within the documentation (which is run as a test), but do lack proper unit tests. (even just smoke screens would be good enough for now)

I am personally in favor of using pytest-matplotlib, but that is low priority

darienmorrow:
@untzag as of yet I haven't thought about how to populate the titles. I am currently thinking about how `mpl_2D` used to have titles that reflected the non-displayed axes dimensions.

darienmorrow:
The relevant `wt2` code is below. 
https://github.com/wright-group/WrightTools/blob/a6ff42f2a36f12a92d186a9532f6ec4cfd58d3c0/WrightTools/artists.py#L2383-L2399

The overlines seem to rely on the notion of `constants`

https://github.com/wright-group/WrightTools/blob/a6ff42f2a36f12a92d186a9532f6ec4cfd58d3c0/WrightTools/artists.py#L2269-L2272

I assume `wt3` no longer has this notion?

ksunden:
Yeah, constant's are not _really_ a thing int wt3, Or rather truly constant constants just have a shape of `(1,) * ndim` (i.e. `size` of `1`).

Some interesting things _could_ be done with any variable that has fewer than two non `1` entries in their (chopped) shape. The question remains whether we *should*. 

Also, there could be some things which are _constant_, but still have `size>1`, but just have repeated numbers (with the added bonus of noise, perhaps). My inclination is to ignore those and put the onus on the from method to collapse where appropriate, rather than trying to guess tolerances or having to check all elements.

ksunden:
With the reintroduction of constants, that wt2 code should be pretty close to working, should be `constant.value` rather than`constant.points` (which gives the mean)

Also may consider converting all constants of the same "kind" to the axis units (not just energy, and instead of doing an identity check on the units themselves. I would perhaps ignore units of `None`, still, because that can potentially be filled with units that are incongruant, and annoying, especially when doing things like plotting motor axes

ksunden:
We seem to be encountering a race condition in our tests where two tests are attempting to use the same data file at the same time. This has happened on roughly 50% of the test runs since the artists tests were added here. It only affects travis, as travis runs with `pytest-mp` to speed up tests.

I have only seen it on Python 3.6, but that may just be where the probability distributions lie with the current test suite.

WrightTools/artists/_quick.py
```diff
@@ -286,6 +298,30 @@ def quick2D(
...
+                # x axis
+                if xaxis.units_kind == constant.units_kind:
+                    constant.convert(xaxis.units)
+                    plt.axvline(constant.value, color="k", linewidth=4, alpha=0.25)
+                # y axis
+                if yaxis.units_kind == constant.units_kind:
+                    constant.convert(yaxis.units)
+                    plt.axhline(constant.value, color="k", linewidth=4, alpha=0.25)
         # colorbar
         cax = plt.subplot(gs[1])
```
ksunden: I think we should put `plt.sca(ax)` after plotting the colorbar so that, should users wish to modify the plot (only particularly useful for single plot calls) they can do so naturally.

That said, these plots are explicitly not for publication quality figures.
One use case I can imagine this being valuable for is to set xlim and ylim in a couple different calls to get the plot area to be the same, while still wanting the quick features like the colorbar and overlines.
A potential additional change along the same lines is to return the axis object when not autosaving (currently returns empty list)

ksunden:
The code looks quite good.
My only reason to hold off on approval is Documentation (and perhaps the `sca` comment above)

darienmorrow:
We should also address #684 in this PR.

WrightTools/artists/_quick.py
darienmorrow: @ksunden I leave you to make these changes if you desire. I am indifferent.

darienmorrow:
Example documentation 
[docs.txt](https://github.com/wright-group/WrightTools/files/2413285/docs.txt)

ksunden:
LGTM shows two new errors for `if not constant.units == None`, this should be `if constant.units is not None`

PR #730, opened 2018-08-28
Join Rewrite

### LINKED ISSUE #503: Recover join overlap methods
```method ({'first', 'sum', 'max', 'min', 'mean'} (optional)) – The method for how overlapping points get treated. Default is first, meaning that the data object that appears first in datas will take precedence.```

Adding 'last', as that is current behavior anyway

This functionality is brought to you by numpy indexing: great power, you know the rest

### LINKED ISSUE #504: Join fill value
Ensure that join has (I believe) `np.nan` as the fill value by default

### LINKED ISSUE #592: joining 3D data objects throws a shape mismatch error.
Consider the below with the attached [files](https://github.com/wright-group/WrightTools/files/1960605/Archive.zip):
```
    import WrightTools as wt
    p = 'short.data'
    d1 = wt.data.from_PyCMDS(p)
    p = 'long.data'
    d2 = wt.data.from_PyCMDS(p)
    ds = [d2, d1]
    d = wt.data.join(ds)
```
This makes two data objects:
```
  axes: ('d2', 'w2__e__wm', 'w1')
  shape: (12, 36, 36)
```
and
```
  axes: ('d2', 'w2__e__wm', 'w1')
  shape: (2, 36, 36)
```
Will note that the `d2` axis has a shape like `(12, 36, 36)`,

The joining operation throws an error.
```
  File "/Users/darienmorrow/source/WrightTools/WrightTools/data/_join.py", line 117, in join
    vals[wt_kit.valid_index(new_idx, old.shape)] = old[:]

ValueError: shape mismatch: value array of shape (2,36,36) could not be broadcast to indexing result of shape (1,1,36,36)
```
I think this may be related to #587

### LINKED ISSUE #719: META join
A place for collecting all of the join thoughts.

Things that join should probably be able to do:
- [x] Join only the shared channels and variables while discarding the rest
- [x] Join overlapping and grid-aligned data
- [x] Join overlapping and non-grid aligned data
- [x] pass the test of `join(x, y, 'first')` being the same as `join(y, x, 'last')` 
- [x] Join 1D slices with 2D planes
- [x] Join 2D planes with 3D boxes
- [x] Join non-overlapping data, grid-aligned data 
- [x] Join non-overlapping, non-grid-aligned data
- [x] Join overlapped and non-overlapped boxes
- [x] Join 3+ data of varying dimension all at the same time
- [x] Pass tests for all of the above
- [x] Join all split tests. Including: http://wright.tools/en/development/auto_examples/split.html#sphx-glr-auto-examples-split-py 

Things that join should maybe be able to do:
- [ ] Join n dimensional data with m dimensional data intersecting at arbitrary skew

### PR DESCRIPTION
**Pull request by ksunden**
Left to do:

- [x] Write 3D tests (closes #592)
- [x] Add tests which address split then rejoin behavior (#719)
- [x] Rewrite to pass tests (closes #719) [This may need subheadings at some point]
- [x] Ensure fill value (closes #504)
- [x] Overlap handling (closes #503)
- [x] Tolerance for merging points (currently hard coded)
- [x] Ensure working for integer dtype
- [x] Update docstring
- [x] Update RTD pages / gallery example

@untzag @p770193 @darienmorrow:
Please feel free to add to this list as you see fit.
Focus is on tests first, then implementation.

For now tests focus primarily on shape, but do test values in many places as well.

Also note, This WILL fail CI until implementation phase, that is expected. I do not intend at this point to mark xfail/skip unless there are test cases which we deem okay to fail when we look to merge.

### CODE REVIEW

**Comment by ksunden** [comment]
All tests are currently passing locally!

Still want to add tolerance kwargs, need to update docs and such. But should be fully usable.

**Comment by ksunden** [comment]
There is a problem that arises from the following test:

```python
data = wt.Data()
data.create_variable("x", np.linspace(-5, 5, 10)[:, None])
data.create_variable("y", np.linspace(-5, 5, 10)[None, :])
data.create_variable("z", np.exp(-data.x[:] ** 2) * np.exp(-data.y[:] ** 2))
data.create_channel("zz", np.exp(-data.x[:] ** 2) * np.exp(-data.y[:] ** 2))
data.transform('y', 'x')
data.print_tree()
j = wt.data.join([data])
```
```
Traceback (most recent call last):
  File "join.py", line 16, in <module>
    j = wt.data.join([data])
  File "/home/kyle/venvs/source/WrightTools/WrightTools/data/_join.py", line 133, in join
    vals[wt_kit.valid_index(new_idx, new.shape)] = old[:]
ValueError: shape mismatch: value array of shape (10,10) could not be broadcast to indexing result of shape (1,10)

```

The problem stems from:
```python       
            for variable_name in vs.keys():
            p = data[variable_name][:][np.newaxis, ...]
            arr = out[variable_name][:][..., np.newaxis]
            i = np.argmin(np.abs(arr - p), axis=np.argmax(arr.shape))
            new_idx.append(i)
```
And the fact that the resultant `p` is not being subtracted from the new axis.
It is, however, crucial that the end result `new_idx` be broadcast, and not simply flat lists (see https://docs.scipy.org/doc/numpy/reference/arrays.indexing.html#advanced-indexing)

My first idea was to add `points`, but that ended up being tricky, need to think it through a bit more.
I also tried (and committed) using boolean masks instead of advanced indexing (which worked in this test), but that proved to have even more challenges, and thus I reverted that commit.

This test case should be added, as the implementation should not falter even when a single data is given, and I believe this to be indicative of larger problems that are not caught by the existing tests.

WrightTools/data/_data.py
```diff
@@ -531,7 +531,9 @@ def convert(self, destination_units, *, convert_variables=False, verbose=True):
                         )
         self._on_axes_updated()
 
-    def create_channel(self, name, values=None, shape=None, units=None, **kwargs) -> Channel:
+    def create_channel(
+        self, name, values=None, shape=None, units=None, *, dtype=None, **kwargs
```
ksunden: I added the `*` for dtype to be kwarg only.
I chose that location for strict backwards compatibility reasons.
That said, given that shape is ignored if values are given, I think it is nonsensical to have shape (or units, for that matter) be accessible only when an ignored (or default) value is given.

Therefore, I move that we place the `*` between values and shape, maintaining the ability to simply pass the values, but making the others keyword only.

Same applies to `create_variable` below

WrightTools/kit/_leastsq.py
```diff
@@ -1,4 +1,4 @@
-"""Interpolation tools."""
+"""Least-square fitting tools."""
```
ksunden: Any particular reason to add this to this PR?

It's a small enough change that now that it's done, not worth reverting, but it is unrelated.

WrightTools/data/_join.py
```diff
@@ -47,13 +71,20 @@ def join(datas, *, name="join", parent=None, verbose=True) -> Data:
         A new Data instance.
     """
     warnings.warn("join", category=wt_exceptions.EntireDatasetInMemoryWarning)
-    # TODO: fill value
+    if isinstance(datas, Collection):
+        datas = datas.values()
     datas = list(datas)
+    if not isinstance(atol, collections.Iterable):
```
untzag: consider `hasattr(atol, '__iter__')`

WrightTools/data/_join.py
```diff
@@ -93,45 +148,122 @@ def from_dict(d, parent=None):
...
+        count[item_name][valid_index] += 1
+        new[:] += vals
+
     for data in datas:
         new_idx = []
+        transpose = []
+        slice_ = []
         for variable_name in vs.keys():
-            p = data[variable_name][:][np.newaxis, ...]
+            # p is at most 1-D by precondition to join
```
untzag: Where is this tested for, and what exception is raised otherwise?

WrightTools/data/_join.py
ksunden: https://docs.python.org/3/library/collections.abc.html#collections.abc.Iterable

This check subsumes the `hasattr` check, though this does raise the point that the Iterable class has moved to `collections.abc` rather than just `collections` (though it is still available there in current python, I think it may be slated for removal in 3.8)

ksunden: the other option is to try/except `iter(atol)`

WrightTools/data/_join.py
ksunden: https://github.com/wright-group/WrightTools/blob/e7f14ed110301d3f440b68ffce014adadb2f8752/WrightTools/data/_join.py#L119

It is actually enforced by the structure of the `from_dict` It will not raise an error, but it will flatten the array (and combine with atol/rtol)

ksunden: So you CAN hand join multidimensional axes, but you will get 1D out

ksunden:
Please advise if there are particular functions that should be exported to kit, the internal functions I define are hyper specific, though there is probably some functionality that is useful elsewhere (default atol/rtol?, the check for floating point type?)

At this point, that kind of reorganizing I think should be a separate PR, or as part of the implementation of a thing that also uses the same code.

ksunden:
If you are satisfied with the changes I made since your last review

PR #735, opened 2018-09-10
ENH: Constants

### LINKED ISSUE #734: constants

ksunden:
closes #734 

To do:
- [x] interface to set constants that exist (analogous to transform for axes, but may want to make adding and subtracting individual items the primary way)
- [x] Add constants in chop/split where appropriate
- [x] Update constants in remove/rename variable
- [ ] ~Add constants in relevant from methods (perhaps to be broken up into separate PRs for each)~
- [x] Floating point representation (currently default number of decimals)
- [x] Tests
- [x] Documentation

No longer planning on adding constants in from_methods, those will come when they are useful/have good heuristics for what to add.


ksunden:
Any opinions on how removing variable contained within constants should behave?

Options are:
- remove the constant
- raise an error, preventing removal (as happens with axes)

p770193:
If I understand constants correctly, they are not as fundamental to the data object as axes are, so I don't think an error is necessary.  Nonetheless, I can imagine users removing these variables inadvertently.  Perhaps allow users to remove, but raise a warning?

ksunden:
Any opinions on how Join should handle constants? (Need to merge from master to make those changes if any are required)

While I haven't tested, I think join would currently inherit  constants from the first data's attrs.
options include:

- Drop all constants
- Drop constants if the variables involved are in the axes that are being joined over
- Include all the constants and maybe have some nonsensical ones
- leave it as is, and see if anybody has issues/desired behavior from experience using

PR #749, opened 2018-09-19
BUG: Parent attribute not present past items in root group

ksunden:
Tricky little subtle mistake that was not previously caught by tests or users, but was actually just plain wrong.

Our inheritance structure meant that `__getattribute__` was finding the correct method, but an error was raised (not an `AttributeError`, but rather a `KeyError`)

Unless python docs are incorrect, an `AttributeError` must have been raised by having the finally clause on this try/except on `_parent`, as the python docs indicate that `__getattr__` is _only_ called when `__getattribute__` raises `AttributeError`

This confusion led to obfuscation of where the problem occured.
I was able to track it down by explicitly calling `col.__getattribute__('parent')` rather than letting it go through it's normal course of method calls.

### DIFF
--- a/WrightTools/_group.py
+++ b/WrightTools/_group.py
@@ -231,7 +231,9 @@ class Group(h5py.Group, metaclass=MetaClass):
         except (AssertionError, AttributeError):
             from .collection import Collection
 
-            key = posixpath.dirname(self.fullpath) + posixpath.sep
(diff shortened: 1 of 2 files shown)


ksunden:
Appveyor failure is due to matplotlib being mid release. Python 3.6 wheels are not out, so it is attempting to build from source, and failing because libraries are not installed.

PR #751, opened 2018-09-28
WIP: Plotting for lower rank channels

### LINKED ISSUE #536: Plotting for lower rank channels
I will explain by way of example:
I have a data object whose shape is (10,100,1000). Call these dimensions `w1`, `w2`, ans `z`
Within this data object, I may have channels which depend on all three, but may also have channels which depend on only `w1` and `w2`, for instance. These channels would be represented by an HDF5 dataset with shape (10,100,1).

I would like to be able to `data.transform('w1', 'w2')` and call `ax.pcolor(data, channel=ch)`

I cannot do this currently as the arrays passed to matplotlib are 3D arrays (though one dimension is 1)

ksunden:
Closes #536

### DIFF
--- a/WrightTools/artists/_base.py
+++ b/WrightTools/artists/_base.py
@@ -108,13 +108,13 @@ class Axes(matplotlib.axes.Axes):
 
     def _parse_plot_args(self, *args, **kwargs):
         plot_type = kwargs.pop("plot_type")
-        if plot_type not in ["pcolor", "pcolormesh"]:
+        if plot_type not in ["pcolor", "pcolormesh", "contourf", "contour"]:
             raise NotImplementedError
         args = list(args)  # offer pop, append etc
         dynamic_range = kwargs.pop("dynamic_range", False)
         if isinstance(args[0], Data):
             data = args.pop(0)
-            if plot_type in ["pcolor", "pcolormesh"]:
+            if plot_type in ["pcolor", "pcolormesh", "contourf", "contour"]:
                 ndim = 2
             if not data.ndim == ndim:
                 raise wt_exceptions.DimensionalityError(ndim, data.ndim)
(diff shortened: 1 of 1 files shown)


darienmorrow:
I have been trying out this PR. I expected to be able to take a 2D dataset `[w1, wm]` collapse along one dimension `d.collapse('w1', method='average')` and then plot with quick1D `wt.artists.quick1D(d, axis='wm', channel='signal_w1_average')`. This all works except for one small problem. quick1D creates many identical images---as many images as the collapsed axis size. I think that this is technically working as intended, but I think it may not be working as we ideally want it to.

In short:
- Plotting of lower rank channels works 
- But quick1D cycles through the chop of the full dataset.

I attach the code and data file by which I found this issue. 
[Archive.zip](https://github.com/wright-group/WrightTools/files/2468881/Archive.zip)

untzag:
@darienmorrow I believe that @ksunden only intends to support quick2D

ksunden:
@untzag, 1D plotting is supported, the changes affect the underlying `plot`, `pcolor`, `contour`, etc methods, not the quick methods. Here, there just needs to be some logic added to assess that the desired shape is lower dimensional, and to prevent `chop` or `quickXD` from producing many outputs along axes that are collapsed in the original channel. I have not yet determined whether this logic should go in chop (adding kwarg for channel, or axes to not iterate, perhaps, but not a huge fan of adding kwarg) or in the quick methods themselves (passing an `at` kwarg to chop, or being selective on the main plotting loop, lean former, but need to think about how it needs to be implemented). 
Current leaning is `at` kwarg, as that does not open up any (new, at least) potentially difficult to maintain capabilities on `chop` that could well have unintended behavior. I also would prefer to limit what comes out of `chop` rather than try to pick out of the list, which seems more fragile.

ksunden:
Test failures appear to be due to using an alpha version of `coverage.py` for some reason...

PR #764, opened 2018-10-29
BUG: axis -> const in __init__

ksunden:


ksunden:
Not sure why travis is not reporting success, it does upon going to their site, ignoring and merging.

PR #770, opened 2018-11-06
Issue 713

p770193:
@ksunden thanks for the catch.

Currently no tests for making sure metadata is imported correctly--should I include that in this PR?


untzag:
`coverage` is failing on the push... :thinking: 

calling all @ksunden s

PR #771, opened 2018-11-07
REL: Add Zenodo Metadata file

ksunden:
Questions for reviewer:

Does the order of the creators list match your expectation?

Should @kjczech have an ORCID?


ksunden:
Merging, we can add, @kjczech's ORCID at a later date if he confirms that he has one.

PR #775, opened 2018-11-12
ENH: Add temperature units, including from_Cary

ksunden:
This allows temperature units in WrightTools data objects, as well as the proper parsing of temperature dependent Cary csv files (tested on a data file which I have not sought permission to share as of yet).

### DIFF
--- a/WrightTools/collection/_cary.py
+++ b/WrightTools/collection/_cary.py
@@ -62,7 +62,7 @@ def from_Cary(filepath, name=None, parent=None, verbose=True):
         name = "cary"
     # import array
     lines = []
-    with open(filepath, "r") as f:
+    with open(filepath, "r", encoding="iso-8859-1") as f:
         header = f.readline()
         columns = f.readline()
         while True:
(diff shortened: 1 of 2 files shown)


ksunden:
@untztag, what do you mean by "symbol support"?

In our off-github conversation, I was under the impression that you were against adding redundant entries to units with the unicode ° symbol to the units dictionary for temperature.

The units do render symbols in LaTeX formating, (`r"^\circ"`), though not for "K", as that is not a "Degree" (arguably the same should be said for Rankine, NIST says so, but many other places retain the °)

ksunden:
I went with "deg_X" for simplicity when typing (though I have my keyboard set up to allow fairly easy ° with <Alt-K>, most non-MacOS people don't)(I use the Macintosh layout on Linux because I got used to the alt codes from my OSX days)

untzag:
add a case to the `get_symbol` function, for label construction

what you are seeing now is fallback behavior

to my vote, seeing the latex formatted °C etc would be preferable

untzag:
also it would be really great to have tests, even nonsense / obfuscated data would be okay

ksunden:
I'll note, we currently have only a smokescreen for units tests right now, though we do test it out in the test for `convert`

ksunden:
Ah, On that front, I agree that I would like one, but need to either seek permission to share such a data file, or collect/fake a dummy one. Would prefer a real one (in part because I don't want to ever be writing to a test of a data file that _couldn't_ be created by the instrument because _I_ typed something wrong.)

ksunden:
I will raise an issue for testing Cary with non-wavelength axis units upon merging

PR #776, opened 2018-11-12
ENH: Gradient as a method of data

### LINKED ISSUE #134: data.gradient
[numpy.gradient](https://docs.scipy.org/doc/numpy/reference/generated/numpy.gradient.html) is a fantastic little method

It would be really nice to have a method `data.gradient` that wraps the numpy method for our pretty little channel objects.

ksunden:
Closes #134

### DIFF
--- a/WrightTools/data/_data.py
+++ b/WrightTools/data/_data.py
@@ -432,11 +432,58 @@ class Data(Group):
             print("chopped data into %d piece(s)" % len(out), "in", new_axes)
         return out
 
+    def gradient(self, axis):
+        """
+        Compute the gradient along one axis.
+
+        New channels have names ``<channel name>_<axis name>_gradient``.
+
+        Parameters
+        ----------
+        axis : int or str
+            The axis to differentiate along.
+            If given as an integer, the axis in the underlying array is used,
+            and unitary spacing is assumed.
+            If given as a string, the axis must exist, and be a 1D array-aligned axis.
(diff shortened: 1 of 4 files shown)


ksunden:
As far as multidimensional goes, It will work, as the underlying gradient does, though only one gradient at a time, as that is how `np.gradient` works. Current implementation does _not_ allow specifying of axis positions without specifying an axis by it's string. This may be revisited in the future, but I couldn't think of a good reason to do it without having the axis, so I made the call to be simpler for now. I kept the behavior mirroring `collapse` when given an integer. In this case, it assumes unity spacing, and thus may be off by a scalar for equally spaced data, and incorrect for data that has nonuniform spacing. Official recommendation is to use your axes unless you absolutely know what you are doing (or do not care about the value, just the shape).

In the future, we may wish to add the `*varargs` arguments from the numpy method to enable this functionality. 

We may also wish to accept a tuple for `axes` or `None` and then create multiple new channels. For now, I decided that was complex enough that I wasn't sure it would be unambiguous, so I went simpler for first iteration. When we have multidimensional datasets for which the gradient is meaningful in multiple dimensions, we shall see if we miss this feature.

There is also no way to prevent uninteresting channels from computing the gradient currently. This may be a kwarg in the future, if it is desired.

untzag:
I would actually prefer that the gradient be taken for one channel at a time

I don't think it's typically desired to take the gradient of all channels at once

PR #783, opened 2018-11-13
ENH: Use natural name rather than full path, remove '/chop000/'

ksunden:
With constants being in the title now, this serves even less purpose than it used to, and is, if anything distracting from the purpose of quick1D, in my opinion.


untzag:
does this actually work?

by my best understanding, the name of the _channel_ is never `chop000`, that's the name of the data object...

ksunden:
name is the full path, as inherited from h5py

PR #785, opened 2018-11-13
ENH: Add Figure title with data name

ksunden:


darienmorrow:
Is it useful to add a kwarg which suppresses the addition of the data name to the title?

darienmorrow:
Will note that the kwarg is useful when the name of datasets is not useful to the end user. For instance, in the documentation gallery the [quick 1D example](http://wright.tools/en/stable/auto_examples/quick1D.html) will output 
image
I don't like the `_066`.

ksunden:
I'll note that that example in particular is more indicative of `from_Kent` giving an unhelpful default name. (It just takes the filename [sans extension] of the first file it finds) 

This could be changed in the example by adding the `name=LDS821_TRSF`

untzag:
I agree with @ksunden --- with the quick artists, the less configuration options the better

I'm loathe to add kwargs unless there is a very good reason to, and in this case I think it's just not quite good enough

easy to add later if riots occur

PR #786, opened 2018-11-15
ENH: Implementation of moment

### LINKED ISSUE #617: moments as method of data
fit.py is being removed

ksunden:
Closes #617 

Still lots to do, but I believe it works:

- [x] Extend so that moment kwarg can accept iterable of moments to calculate
- [x] Update Documentation (including docstring, currently copied from collapse.)
- [x] Write Tests
- [ ] Make appropriate changes to collapse that result from deprecating integrate there (may break off to separate PR)
- [ ] ~Perhaps do some special casing for moment 1 because weighted average computation could be more efficient than computing the integral, result is the same~ (about 25% faster with 1000 points, but difference is microseconds. Keep it simple for now. Savings are larger as n grows, but also requires increasing dimensionality of x, it seems. Basically, wait until this is a bottleneck [it may be for `attune`, honestly])
- [ ] ~Look into allowing users to specify which _kind_ of moment they want ('central', 'normalized', 'raw')~ (Current leaning: wait until need to do so is expressed, keep it simple until then)
- [x] Ensure proper handling of NaN, probably wish to ignore nans. perhaps require that the axis has no nans, and then call `nan_to_num` on the channel's numpy array used for computation


ksunden:
Test failures are python 3.5, see #798

ksunden:
In order to make review simpler, I am forgoing a gallery example here, but do plan on adding one as a separate PR.

The same is true of changes to collapse, those will be a separate PR

PR #788, opened 2018-11-15
clip out negative pixels in label_delay_space

untzag:
in this case, the grey pixels are just distracting and don't contribute to the purpose of the example

note to reviewers:
:warning: this code is not tested---please render before merging :warning: 

thanks, sorry


ksunden:
The special kwarg is `'value'`, not `'number'`

PR #791, opened 2018-11-18
set_fig_labels enhancements

### LINKED ISSUE #139: set_fig_labels does not write over existing axis labels that are not on the left or bottom edges of a figure

### LINKED ISSUE #765: Add kwargs to set_fig_labels to allow non-first-column and non-last-row to be the labeled row and column
example syntax: `wt.artists.set_fig_labels(col=0, row=-1)`

untzag:
resolves #139 
resolves #765

### DIFF
--- a/WrightTools/artists/_helpers.py
+++ b/WrightTools/artists/_helpers.py
@@ -936,10 +936,10 @@ def set_fig_labels(
         yticks. If False, ticks are hidden. Default is None.
     title : None or string (optional)
         Title of figure. Default is None.
-    row : integer (optional)
-        Row to label. Default is -1.
-    col : integer (optional)
-        col to label. Default is -1.
+    row : integer or slice (optional)
+        Row to label. Default is -1. If slice, step is ignored.
+    col : integer or slice (optional)
+        col to label. Default is 0. If slice, step is ignored.
     label_fontsize : number (optional)
         Fontsize of label. Default is 18.
     title_fontsize : number (optional)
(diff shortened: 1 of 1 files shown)


WrightTools/artists/_helpers.py
```diff
@@ -946,32 +952,33 @@ def set_fig_labels(
...
-                ax=ax, ylabel=ylabel, xticks=False, yticks=yticks, label_fontsize=label_fontsize
-            )
-        elif ax.is_last_row():
-            # bottom row
-            set_ax_labels(
-                ax=ax, xlabel=xlabel, xticks=xticks, yticks=False, label_fontsize=label_fontsize
-            )
-        else:
-            set_ax_labels(ax=ax, xticks=False, yticks=False)
+        if True: #row_start < ax.rowNum < row_stop and col_start < ax.colNum < col_stop:
```
darienmorrow: I think this is the correct set of truth statements. However, I don't think they are implemented fully below. For instance, a column after `col_stop` still has its ylabel changed. This problem also exists for rows before `row_stop` being changed. Consider the below example:
```
import numpy as np
import matplotlib.pyplot as plt
import WrightTools as wt

fig, gs = wt.artists.create_figure(nrows=2, cols=[1,1,'cbar'])

indxs = [(row, col) for row in range(2) for col in range(2)]

for i, indx in enumerate(indxs):
    ax = plt.subplot(gs[indx])
    ax.set_ylabel('y')
    ax.set_xlabel('x')
    
cax =   plt.subplot(gs[:,-1])  
cax.set_ylabel('cax')

row = slice(0, 1, None)
col = slice(0, 1, None)
wt.artists.set_fig_labels(row=row, col=col)
```
I expect the last column (`cax`) to not be touched w.r.t. its yaxis. It however is has its ticks and labels removed. 
image

untzag: try your code again---it should work this time

PR #796, opened 2018-11-24
clip data for level example

untzag:
I think the undeflow behavior is distracting in this example


darienmorrow:
I agree with @untzag

ksunden:
Finally looked at the actual example, and I too agree

PR #797, opened 2018-11-24
clip data for fill_types example

untzag:
ready to merge, although I'm not exactly pleased with current behavior of `contourf`


darienmorrow:
When I run `fill_types.py` on my OSX machine I get the following stdout 
```
cols recognized as v0 (19)
/Users/darienmorrow/source/WrightTools/WrightTools/_group.py:85: UserWarning: 'source' not included in attrs because its Type (<class 'list'>) cannot be represented
  warnings.warn(message)
data created at /var/folders/cx/nn1m8vzn6n350jhz_2z33hqw0000gn/T/4p04xedu.wt5::/
  axes: ('d1', 'd2')
  shape: (21, 21)
channel ai0 leveled along axis 0
```
The last four lines resemble that found on the [old documentation](http://wright.tools/en/stable/auto_examples/fill_types.html#sphx-glr-auto-examples-fill-types-py). But the first line is new. Is this worrisome to anyone?

ksunden:
@darienmorrow I'd have to check, but this is related to #811, gahh, apparently `isinstance(str, os.PathLike)` is False.... I thought str was special cased there, I'll submit a patch to fix

PR #800, opened 2018-11-27
ENH: Use natural_name setter to rename item in hdf5 file

### LINKED ISSUE #417: api for renaming group
need api to cleanly rename objects derived from `Group` (`Collection` and `Data`)

we need to be careful because renaming a `Group` implies changing the `fullpath` of all objects that it contains---this is annoying because of the way we use the `instances` class attribute

ksunden:
Closes #417

### DIFF
--- a/WrightTools/_dataset.py
+++ b/WrightTools/_dataset.py
@@ -4,7 +4,6 @@
 # --- import --------------------------------------------------------------------------------------
 
 
-import posixpath
 import collections
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


WrightTools/_dataset.py
```diff
@@ -87,8 +86,7 @@ def __init__(self, *args, **kwargs):
     def __new__(cls, parent, id, **kwargs):
         """New object formation handler."""
         fullpath = parent.fullpath + h5py.h5i.get_name(id).decode()
-        if fullpath.startswith("//"):
-            fullpath = fullpath[1:]
+        fullpath = fullpath.replace("//", "/")
```
untzag: :clap:

Issue #365, opened 2017-11-07
cache


untzag:
`wt.cache[key] = value` where value instance of Group

wt5 file stored in OS-specific cache

`__getitem__` returns copy of object, original is never modified unless user overwrites that key

in future, may populate cache with content from the internet ✨ automagically ✨

ksunden:
Is this a system we want now, or should this be moved to future? Last I knew, @kjczech was the one who desired this functionality sooner rather than later.

untzag:
honestly, I forgot about this plan

I lean towards future

@kjczech, what say you?

untzag:
at the moment I feel like this probably should not be built into WrightTools itself... considering closing if no objection

untzag:
out of scope

PR #801, opened 2018-11-28
ENH: np.DataSource for transparent compression and remote files

ksunden:
- [x] Implementation (those that use genfromtxt needed no changes)
- [x] Document from a users perspective (add to docstrings, mention in quickstart, maybe a few other places in human generated rst files)
- [x] Document from developers perspective (Edit page on writing a from method with explicit instructions on how to get it to work.)
    - [x] Any call to `open` should be `ds = np.DataSource(None); f = ds.open(fname, 'rt');...;f.close()`
- [x] Write tests for behaviour, particularly remote files. Already have made files compressed, may desire to compress all ascii text files, which would then want at least a few tests of the uncompressed file for good measure.
- [ ] ~Decide whether to include some remote files in datasets (could go so far as making only remote), could also make those files persistent and add code to look for previous downloads.~ Not this PR, but keep in mind for the future, perhaps the near future with Kohler2017 data?, Regardless, that engineering is not part of this already pretty hefty PR.


ksunden:
Perhaps we should use bz2 instead of gz for one of the solis datasets, It would ensure that both work

ksunden:
Kind of reminds me of #365 in a way. not the same, but perhaps some of the same ideas.

If we ever wish to revisit the idea of a cache for datasets, (or possibly in some future iteration of `wt.datasets`) DataSource may provide a way of doing it fairly simply. May consider taking in an optional kwarg in from methods for a DataSource object (or path to create one)... 

It's an idea, but not a fully thought out one at this point.

ksunden:
Test failures are because I'm using pre-release tidy_headers here

WrightTools/collection/_directory.py
```diff
@@ -66,26 +67,26 @@ def from_directory(filepath, from_methods, *, name=None, parent=None, verbose=Tr
...
                         parent=parent,
                         verbose=verbose,
                     )
                 break
         else:
-            if os.path.isdir(os.path.join(path, fname)):
+            if (path / fname).is_dir():
                 if verbose:
-                    print("Creating Collection at", posixpath.join(parent.name, fname))
+                    print("Creating Collection at", pathlib.PosixPath(parent.name) / fname)
```
ksunden: perhaps should be PurePosixPath here...

ksunden:
I apologize to the review for the volume of change in this PR, but it is mostly repetitive changes, so that helps.

setup.py
```diff
@@ -44,7 +44,7 @@ def read(fname):
         "python-dateutil",
         "pytz",
         "scipy",
-        "tidy_headers>=0.3.0",
+        "tidy_headers>=1.0.0",
```
darienmorrow: This will conflict with #813

Issue #677, opened 2018-07-25
Documentation for create_figure should reflect subclassed returns


ksunden:
https://github.com/wright-group/WrightTools/blob/cecddd9025678f390b55df0498bd27530faa37c4/WrightTools/artists/_helpers.py#L291

Currently we say we return matplotlib.figure.Figure/matplotlib.gridspec.GridSpec

We actually return WrightTools.artists.Figure and WrightTools.artists.GridSpec objects, which are subclasses of the matplotlib versions which have knowledge of WrightTools.artists.Axes, which know how to plot data objects. (Granted, WrightTools.artists.GridSpec currently does exactly nothing other than change the namespace)

PR #803, opened 2018-11-28
DOC: Custom figure documentation and example

darienmorrow:
- [x] link from `data.constants` to this example
- [x] Provide prose walking through example


ksunden:
#677 is primarily about docstrings, which I don't think I want to touch here. That said, I will mention the same Ideas in this page.

PR #805, opened 2018-11-28
DOC: TEST: heal 

darienmorrow:


tests/data/heal.py
```diff
@@ -0,0 +1,48 @@
...
+
+import WrightTools as wt
+
+import matplotlib.pyplot as plt
+
+
+# --- test ----------------------------------------------------------------------------------------
+
+
+def test_1():
```
ksunden: Would like a name like `test_heal_gauss` or something

PR #807, opened 2018-11-28
DOC: map-variable example

darienmorrow:


examples/map-variable.py
```diff
@@ -0,0 +1,40 @@
...
+# create data object
+d = wt.data.Data(name="original")
+d.create_variable("x", values=x)
+d.create_variable("y", values=y)
+d.create_channel("z", arr)
+d.transform("x", "y")
+# create new data objects
+pointsx = np.linspace(x.min(), x.max(), 31)  # linear spacing
+pointsy = points = np.logspace(0, .7, 11)  # log spacing
+dx = d.map_variable("x", points=pointsx)  # just linear along x
```
ksunden: Do we wish to demonstrate just passing an int here?

darienmorrow: I am indifferent.

PR #812, opened 2018-11-30
BUG, ENH: Fix some chunkwise operations

### LINKED ISSUE #759: Make clip argument chunkwiseable
Clip fails with chunked data if you pass an array, such that you wish to clip based on different values along some given axis, something that is useful for, e.g. clipping to the value at a given index of a frame in a higher dimensional scan

Could also apply to other similar chunked functions

ksunden:
Clip now accepts arrays as min, max, replace (Closes #759)

Fix argmin/max to maintain nan-awareness if a full chunk is nan

### DIFF
--- a/WrightTools/_dataset.py
+++ b/WrightTools/_dataset.py
@@ -192,14 +192,13 @@ class Dataset(h5py.Dataset):
                 arr = dataset[s]
                 amin = np.nanargmax(arr)
                 idx = np.unravel_index(amin, arr.shape)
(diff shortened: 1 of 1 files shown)


ksunden:
Turns out, arg min/max were broken entirely for chunked data _and_ the test was incorrect

ksunden:
Remaining test failure is `get_zenith`, which calls `argmax` and I believe, but should be verified, that the test itself is incorrect

PR #816, opened 2018-12-03
BUG: respect channel signed header in data

ksunden:


ksunden:
Does this need backporting?

PR #817, opened 2018-12-04
[WIP] small potatoes

darienmorrow:


darienmorrow:
I am done frying the small potatoes. 

The paper builds, however, the long list of citations in the impact section don't quite render correctly w.r.t. author truncation. Moreover, Kohler2018 doesn't appear in the list, but does appear in the reference list but with a strikethrough as an author list [paper.pdf](https://github.com/wright-group/WrightTools/files/2644361/paper.pdf).

 I am unable to figure out these citation issues. @ksunden do you want to take a stab at them?

ksunden:
The dashes in the author list in the reference section is just the citation style, indicating that the author list is the same as the previous entry, the same is true of the "Kohler, Thompson, and Wright 2017, 2018;" in that long citation list

ksunden:
pandoc-citeproc defaults to Chicago Author Date style

darienmorrow:
Well, I certainly spent way to long this morning trying to solve that non-problem.

ksunden:
All that said, I do find it odd that sometimes it lists only one author et al. others it lists three et al.

At first glace seems to be correlated with number of authors, but not really sure

PR #823, opened 2018-12-07
BUG, MAINT: __dict__.pop --> delattr

ksunden:


ksunden:
This prevents a crash if you have a malformed data object with two of the same axes, and then call anything that tries to update the names (e.g. transform) to fix that

PR #825, opened 2018-12-08
DOC: Split api into separate files

ksunden:
- [x] artists
- [x] collection
- [x] data
- [x] exceptions
- [x] kit
- [x] units

Not going to bother with diagrams (which is already max of two sections on a page)/just going to remove datasets (as it is not documented, and is dynamically generated to a certain degree)

Will likely keep small classes from kit on single page for the class


ksunden:
Close to done, may need to think about TimeStamp class and want to add `:show-inheritance:` in several places

ksunden:
You can play with this at http://wright.tools/en/split_docs/api/

untzag:
@ksunden here are my requested changes

- [ ] remove `artists.Figure.add_subplot` (we don't actually want people using this method)
- [ ] remove `data.Data.keys`, `data.Data.items`
- [ ] remove `data.Data.move`
- [ ] remove `data.Data.pop`, `data.Data.popitem`
- [ ] remove `data.Data.setdefault`
- [ ] remove `data.Data.visit`, `data.Data.visititems`
- [ ] remove `data.Data.ref`, `data.Channel.ref`, `data.Variable.ref`
- [ ] remove `data.Channel.astype`, `data.Variable.astype`
- [ ] remove `data.Channel.len`, `data.Variable.len`
- [ ] remove `data.Channel.fletcher32`, `data.Variable.fletcher32`
- [ ] remove `data.Channel.maxshape`, `data.Variable.maxshape`
- [ ] remove `data.Channel.scaleoffset`, `data.Variable.scaleoffset`
- [ ] remove `data.Channel.shuffle`, `data.Variable.shuffle`
- [ ] remove `data.Channel.resize`, `data.Variable.resize` (these break WrightTools)
- [ ] I cannot click on the `INI` link under the `WrightTools.kit` landing page
- [ ] make it obvious that `kit.Timer` has attributes `end`, `interval`

why is 'return type' so screwy (e.g. data from functions)

thanks for all this cleanup---looks a lot better already

ksunden:
As for return type, that is 100% on the docstring, which I'm leaning towards that being a separate PR.

I do think we should pick one style and stick to it, with my preference being to one that properly links to the object in the api.

ksunden:
I'm slightly unhappy with `:show-inheritance:` behavior, as I do not know how/if we can control what shows up there.

This is not new, but artists.Axes points to `matplotlib.axes._axes.Axes`, not `matplotlib.axes.Axes`, which is where the docs are found.

Similarly, I would like to to show inheritance of Data/Collection and Channel/Variable from h5py, but they link to our (undocumented) `_group.Group` and `_dataset.Dataset`

In all, I think I still like having it, as it adds some context, but would prefer if I could control it a bit more

PR #826, opened 2018-12-08
templates

untzag:
adding issue and PR templates so people don't forget to assign labels, milestones etc

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,13 @@
+### issue TODOs
+
+When opening a `WrightTools` issue, please follow these steps:
+
+1. Search for a related issue. If a related issue exists, consider making a comment or editing that issue rather than spawning a new one.
+2. Give your issue a concise, informative title.
+3. Consider assigning someone to resolve this issue, perhaps yourself. Only assign someone if you are certain that they will follow through quickly.
+4. If possible, apply appropriate labels (don't worry, these can be changed later).
+5. If possible, assign a milestone (don't worry, this can be changed later).
+6. In this text box, describe the issue as clearly as possible.
+7. Delete this TODO list :wink:
+
+Thanks so much!
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,14 @@
+### PR TODOs
+
+When making a PR to `WrightTools`, please follow these steps:
+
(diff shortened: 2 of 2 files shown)


ksunden:
Consider making these lists comments rather than asking the user to delete them

Markdown style comments (a bit long, I had to look it up, but it exists and works on github):
```
[//]: # (This actually is the most platform independent comment
1.
2.
3.
)
```

[//]: # (This actually is the most platform independent comment
1.
2.
3.
)

or 

html style comment
```
<!---
1. a
2. b
3. c
-->
```

ksunden:
Also, I don't know how I feel about asking to assign tags/milestones to issues: 

To me, templates are primarily for irregular/new contributors, and I do not see tags/milestones as their job, it is a job for regular contributors/maintainers (I don't think those without commit bit _can_ do it, and asking them to [even with `if possible`] makes it sound like they should)

I also think tags are something I as a maintainer want to assign, partly as a way of acknowledging the issue and giving some more immediate feedback. Sure, when one of the maintainers raises the issue, I think them assigning tags is 100% appropriate.

Milestones are to me the domain of the release manager, I'm good with maintainers chiming in with their expectation, but the one planning out the release has final say.

ksunden:
Another thing that I think is good to do, which is not included in these templates, is provide a structure to the issue (mostly for bug reports, not feature requests), things like collecting version information, so that you have to ask for example less often:

e.g., from numpy

```markdown

<!-- Please describe the issue in detail here, and fill in the fields below -->

### Reproducing code example:

<!-- A short code example that reproduces the problem/missing feature. It should be
self-contained, i.e., possible to run as-is via 'python myproblem.py' -->

`literal block removed by KFS, including import`

<!-- Remove these sections for a feature request -->

### Error message:

<!-- If you are reporting a segfault please include a GDB traceback, which you
can generate by following
https://github.com/numpy/numpy/blob/master/doc/source/dev/development_environment.rst#debugging -->

<!-- Full error message, if any (starting from line Traceback: ...) -->

### Numpy/Python version information:

<!-- Output from 'import sys, numpy; print(numpy.__version__, sys.version)' -->
```

ksunden:
Part of me wishes we could have multiple issue templates, one for Bug, one for Feature request, possible even some more specific ones for things like `from_method` requests, but to my knowledge, that is not supported on github (it is on gitlab, though...)

ksunden:
I retract my last statement, it _is_ possible to have multiple different templates

See #831, which was simply me messing around with default templates (only actually committed /PRed because I couldn't see a true diff otherwise)

PR #857, opened 2019-01-08
MAINT/TST: remove multiprocessing from tests

ksunden:
Causes pytest to fail with recent update


ksunden:
The tests do run slower without mp, but not as dramatically as you might think.
It is at about 14-15 minutes vs 9-10 with mp. Appveyor is more of a bottleneck anyway (and never had mp), so not usually that big of a deal.

PR #859, opened 2019-01-10
ENH: recover contour in Quick2D (resolves #815)

### LINKED ISSUE #815: Recover contour for Quick2D
https://github.com/wright-group/WrightTools/blob/86ebafa5910fb7614552feec13279f53299e13b6/WrightTools/artists/_quick.py#L312

darienmorrow:


WrightTools/artists/_quick.py
```diff
@@ -309,7 +309,23 @@ def quick2D(
...
+                else:
+                    limit = data_channel.mag()
+                contour_levels = np.linspace(
+                    -limit + channel.null - 1e-10, limit + channel.null + 1e-10, contours + 2
+                )
+            else:
+                if contours_local:
+                    limit = np.nanmax(zi)
+                else:
+                    limit = data_channel.mag()
```
ksunden: Should this be max?

WrightTools/artists/_quick.py
```diff
@@ -309,7 +309,23 @@ def quick2D(
...
+                if contours_local:
+                    limit = channel.mag()
+                else:
+                    limit = data_channel.mag()
+                contour_levels = np.linspace(
+                    -limit + channel.null - 1e-10, limit + channel.null + 1e-10, contours + 2
+                )
+            else:
+                if contours_local:
+                    limit = np.nanmax(zi)
```
ksunden: should this be channel.max()?

WrightTools/artists/_quick.py
```diff
@@ -309,7 +309,23 @@ def quick2D(
...
+                else:
+                    limit = data_channel.mag()
+                contour_levels = np.linspace(channel.null - 1e-10, limit + 1e-10, contours + 2)
```
ksunden: Why are we avoiding a contour at channel.null?

I get avoiding the maximum and the min/max in signed, but why null?

darienmorrow: Were we not to do this, if there is data at (or below) null, then contour would be present and unhelpful.

WrightTools/artists/_quick.py
darienmorrow: Sure.

WrightTools/artists/_quick.py
darienmorrow: I think so. I'll change it.

WrightTools/artists/_quick.py
darienmorrow: TBH, this is how it was written in `mpl_2D` and I just kept the idea.

ksunden: if there's data below null here, it will be present anyway

PR #860, opened 2019-01-15
Move to service instead of involking xvfb directly

ksunden:


ksunden:
This does fix one failure with the newer Travis image, however tests are still failing on 3.6 due to incompatability of skimage and numpy 1.16.0 (they access a private variable that disappeared), this should be fixed with the next skimage release, whenever that occurs.... Or, we can remove the dependency

untzag:
@ksunden should we pin to an earlier version of numpy?

ksunden:
@untzag No, I don't think we should do that, maybe on CI for now, but certainly not in `setup.py`

I'm more interested in removing skimage as a dependency

PR #861, opened 2019-01-16
JOSS: update punctuation to be after period

darienmorrow:
Addresses https://github.com/openjournals/joss-reviews/issues/1141#issuecomment-454588495 . 

I think I got all instances of periods being before citations...


untzag:
thanks so much @darienmorrow

PR #862, opened 2019-01-16
MAINT: Use scipy.signal instead of skimage for downscale

ksunden:
Removes dependency scikit-image

Supercedes #860 (needed for testing on travis)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,6 @@ install:
   - pip install cython
   - pip install .
   - pip install -U pytest "coverage<5"
-before_script:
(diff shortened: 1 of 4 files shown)


ksunden:
We may consider playing with adding parameters to control behavior, in particular, since this is doing some filtering, it can have effects outside of the window of interest, rather than simply averaging the points.

This also changes behavior to preserve all variables, and apply the same decimate averaging, instead of just directly downsampling.

ksunden:
If merged, requires editing the conda-forge recipe on next release, as dependency list is repeated there

PR #863, opened 2019-01-16
MAINT: Downscale convolve

ksunden:
Alternative to #862

### DIFF
--- a/WrightTools/data/_data.py
+++ b/WrightTools/data/_data.py
@@ -15,7 +15,7 @@ import h5py
 
 import scipy
 from scipy.interpolate import griddata, interp1d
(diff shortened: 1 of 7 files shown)


WrightTools/data/_data.py
```diff
@@ -897,10 +897,9 @@ def create_variable(
     def downscale(self, tup, name=None, parent=None) -> "Data":
         """Down sample the data array using local averaging.
 
-        See `skimage.transform.downscale_local_mean`__ for more info.
+        See `scipy.signal.decimate`__ for more info.
```
ksunden: Left over from first rewrite, if we move forward with this one, needs updating. 

Also applies to internal function name

PR #865, opened 2019-01-17
DEP/MAINT: Remove downscale/skimage dependency

ksunden:
Mostly as a way of ensuring that current master works enough to do proper code reviews.

See #862 and #863 for implementations of downscale as it may reappear.


ksunden:
Requires #860

PR #873, opened 2019-02-14
Proof of Concept, using multidimensional axis in moment

ksunden:
Really not sure I like the API here, but the idea is that moment needs to know what axis in the underlying array it is "along", but should work just fine even if the axis itself is multidimensional

See the `attune/poc_mdaxis_moment` for example usage

### DIFF
--- a/WrightTools/data/_data.py
+++ b/WrightTools/data/_data.py
@@ -537,15 +537,20 @@ class Data(Group):
             Set values above/below a threshold to a particular value
         """
         # get axis index --------------------------------------------------------------------------
+        axis_index = None
+        if isinstance(axis, tuple):
+            axis_index = axis[0]
+            axis = axis[1]
         index = wt_kit.get_index(self.axis_names, axis)
-        axes = [i for i in range(self.ndim) if self.axes[index].shape[i] > 1]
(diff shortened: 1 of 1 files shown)


ksunden:
Do not merge, Proof of concept only at this point

ksunden:
Perhaps it would be cleaner to add a new optional kwarg `along` which specifys the array axis. moments needs the "real" axis to get the points still (in unit'd values, though the units don't actually affect the computation) but the dimension which is collapsed could be just one of those, I think, as is implemented here.

ksunden:
What we are currently thinking is adding a kwarg "resultant", which you hand the shape of the resultant channel. This should have exactly one additional "1" compared to the shape of the channel being collapsed, and the axis being collapsed should not have a 1 in that same spot.

`wt.kit.joint_shape` will be a useful method, as you can do things like `joint_shape(d.w1, d.w1_c2)` to get the shape spanned by multiple axes.

ksunden:
An error message from this method should indicate `resultant` and `joint_shape` as potential solutions when invalid calls are made

ksunden:
closing in favor of #883

PR #878, opened 2019-02-28
BUG: Make pcov a numpy scalar

### LINKED ISSUE #868: pcov TypeError in kit._leastsq
In kit._leastsq, if the line 62 if statement is not passed, the consequent else statement makes pcov data type float, triggering"TypeError: 'int' object is not subscriptable" in line 72-73:

72:      try:
73:                error.append(np.absolute(pcov[i][i]) ** 0.5)

Line 74 picks up index out of bound errors, not sure if it was meant to catch the type error.

74:      except IndexError:
75:                 error.append(0.00)

Error is bypassed if I put a 2D array into line 68, but have not spent the time considering what this array should look like.

ksunden:
Fixes #868 

Should we add a test? if so, what can we do to reliably hit this condition?

### DIFF
--- a/WrightTools/kit/_leastsq.py
+++ b/WrightTools/kit/_leastsq.py
@@ -65,7 +65,7 @@ def leastsqfitter(p0, datax, datay, function, verbose=False, cov_verbose=False):
             if cov_verbose:
                 print(pcov)
         else:
-            pcov = np.inf
+            pcov = np.array(np.inf)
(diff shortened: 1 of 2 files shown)


darienmorrow:
First things that comes to mind is to try to fit two variables which are anticorrelated with eachother. Or to fit a single variable which does not change the output of the function.

PR #883, opened 2019-03-20
Implement `resultant` kwarg on moment

ksunden:
Supersedes #837

Allows the user to specify the resultant shape of taking the moment
for when the "x-axis" for the moment computation is itself multidimensional.


ksunden:
Test failure is due to old sphinx, as this was branched off of #873 which was branched before #877, only failure is docs building, no relation to the changes, the PR version passed

ksunden:
I like green checkmarks, so merged master

untzag:
@ksunden do you plan on writing tests for this new kwarg?

ksunden:
Good call @untzag, In writing tests, I discovered that 0th moment doesn't work at the moment (pun intended) The problem is just that we have a sign correction piece that is doing weird things.

I had seen it earlier, but convinced myself it wasn't actually a problem (see https://github.com/wright-group/WrightTools/pull/883/commits/b36c28e6b5aca4037c1e4facdd073cf2dede38fd for the revert of my first attempt).

It seems I was indeed originally correct and it is something we need to address... perhaps the solution is to have a per pixel +/- 1 array something like `np.sign(x[-1] - x[0])` (with thought towards the order, and those indicies being in axis dimension) instead of doing the actual process of sorting, as (if monotonic, as it should be for this method) it is just a sign problem

Issue #890, opened 2019-03-29
Matplotlib deprecation for clim


ksunden:
```
/home/kyle/venvs/source/WrightTools/WrightTools/artists/_base.py:532: MatplotlibDeprecationWarning:
The process_projection_requirements function was deprecated in Matplotlib 3.1 and will be removed in 3.3.
  self, *args, **kwargs

/venvs/source/WrightTools/WrightTools/artists/_helpers.py:687: MatplotlibDeprecationWarning:
The set_clim function was deprecated in Matplotlib 3.1 and will be removed in 3.3. Use ScalarMappable.set_clim instead.
  cbar.set_clim(clim)
```

Two deprecation warnings appear on current matplotlib master.
Pretty easy to fix, though currently there is a regression that I may track down and PR before mpl is released causing a breakage (not just a warning) for the clim, as we pass `set_clim(clim[0], clim[1])`

~~This is fixed by calling `set_clim(clim)` (though the warning still appears for now, have not tried using the suggested alternative yet)~~ Fix submitted upstream see matplotlib/matplotlib#13789 the function signature was incorrect.

ksunden:
the process_projection_requirements is in `add_subplot` of our `Figure` subclass in `_base.py`.
`clim` is in `plot_colorbar` in `_helpers.py`

See also matplotlib/matplotlib#13234

ksunden:
process_projection_requirements is becoming a private method, not going away completely.

While I don't like using private methods which can change more rapidly, we are using it in a subclass already, so private methods are more reasonable. Just need someone to stay on track and make sure things don't break (which means myself).

ksunden:
https://github.com/matplotlib/matplotlib/commit/4ffe21beb8c6afb02f179c3ff40cd487f0f3b7bf

The commit in question

ksunden:
Will require a try/except to support both old and new, but not too bad on the whole, there

anntzer:
Sorry for the change (I did the deprecation), but it looks like your entire `add_subplot` (which I only quickly skimmed thru) could just be (untested)
```
def add_subplot(self, *args, **kwargs):
    kwargs.setdefault("projection", "wright")
    super().add_subplot(*args, **kwargs)
```
and then you don't have to bother with the change?

darienmorrow:
https://github.com/wright-group/WrightTools/blob/10a29b10baf06d7971e21b6568476bfe661ae19f/WrightTools/artists/_helpers.py#L687

[...]

PR #891, opened 2019-03-29
MAINT: Use register_projection to enable WT Axes

ksunden:
Thanks to @anntzer for the suggestion, it worked perfectly (with a
couple of lines needed to actually properly register the class)

This addresses half of the deprecation warnings found in #890

This also enables using our axes class without calling our own `Figure`
class.

```
plt.subplot(111, projection="wright")
plt.plot(data)
```

Question for @untzag/@darienmorrow, Do we wish to document the above
behavior and make the projection name more public facing than it
already is (currently considered implementation detail, especially as it
was not accessible by other means previously).
If so, do we also wish to name it "wright_tools" instead (or some other
variation) to make it more explicitly linked to our library?

### DIFF
--- a/WrightTools/artists/_base.py
+++ b/WrightTools/artists/_base.py
@@ -7,6 +7,7 @@
 import numpy as np
 
 import matplotlib
+from matplotlib.projections import register_projection
 from matplotlib.axes import SubplotBase, subplot_class_factory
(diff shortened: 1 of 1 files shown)


ksunden:
also `plt.axes(projection="wright")`

PR #898, opened 2019-04-09
Do not edit local when print tree

ksunden:
I found that the `wt-tree` command was modifying the files themselves (not sure what was written, just that git told me they were different, and indeed the SHA sums were different)

Since tree _should_ be passive, this PR removes the `edit_local=True` kwarg such that it will create a copy rather than modify the original.

This impacts performance, as it has to do the copy operation, but means that it does not change the file itself, so that's a tradeoff I think we make.

### DIFF
--- a/WrightTools/__main__.py
+++ b/WrightTools/__main__.py
@@ -19,7 +19,7 @@ def wt_tree():
     args = parser.parse_args()
 
     # Create the data/collection object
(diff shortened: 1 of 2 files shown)


ksunden:
For the record, appveyor failure is not real, failure to resolve "github.com"

PR #906, opened 2019-05-22
MAINT: Add explicit file modes to h5py File calls

### LINKED ISSUE #900: h5py deprecation: will default to read-only in 3.0
Currently defaults to "append"

```
/home/kyle/venvs/source/WrightTools/WrightTools/_open.py:56: H5pyDeprecationWarning: The default file mode will change to 'r' (read-only) in h5py 3.0. To suppress this warning, pass the mode you need to h5py.File(), or set the global default h5.get_config().default_file_mode, or set the environment variable H5PY_DEFAULT_READONLY=1. Available modes are: 'r', 'r+', 'w', 'w-'/'x', 'a'. See the docs for details.
```

```
_group.py
156:        if isinstance(filepath, h5py.File):
184:            file = h5py.File(p, "a")
395:        h5py.File(filepath)

_open.py
56:    f = h5py.File(filepath)
```

Looks like only two lines where this needs to be fixed, and the `_open` one may be perfectly fine as "r", as far as I know, that is only used to read attrs to determine which constructor to call.

ksunden:
Closes #900

### DIFF
--- a/WrightTools/_group.py
+++ b/WrightTools/_group.py
@@ -392,7 +392,7 @@ class Group(h5py.Group, metaclass=MetaClass):
                 raise wt_exceptions.FileExistsError(filepath)
 
         # copy to new file
-        h5py.File(filepath)
+        h5py.File(filepath, "w")
         new = Group(filepath=filepath, edit_local=True)
         # attrs
         for k, v in self.attrs.items():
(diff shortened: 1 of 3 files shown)


ksunden:
@ddkohler You should be back to having a big green button (after CI passes, which tests passed locally, so CI should pass), as I did branch off of master with your updated username in CODEOWNERS.

Github is indeed indicating you as a reviewer as code owner, so seems to be correct from this end

PR #907, opened 2019-05-22
Interact 2d key nav

### LINKED ISSUE #733: interact2D:  use arrow keys to navigate sliders
Left and right arrows key presses move slider values

For this to work with multiple sliders, we must assign focus to a slider--visually, focus will be shown by a thicker slider border.

Slider focus can be selected by mouse clicks or tab/shift-tab

ddkohler:
Simple keyboard controls for navigating.  tab/ctrl+tab to alternate between sliders and the plot, arrow keys to adjust the slider values/side plot slices.

Closes #733

### DIFF
--- a/WrightTools/artists/_interact.py
+++ b/WrightTools/artists/_interact.py
@@ -211,7 +211,8 @@ def interact2D(data, xaxis=0, yaxis=1, channel=0, local=False, verbose=True):
     line_sp_y = sp_y.plot([None], [None], visible=False, color=y_color)[0]
     crosshair_hline = ax0.plot([None], [None], visible=False, color=x_color)[0]
     crosshair_vline = ax0.plot([None], [None], visible=False, color=y_color)[0]
-    current_state.xpos = crosshair_hline.get_ydata()[0]
+    current_state.xarg = 0
+    current_state.yarg = 0
     current_state.ypos = crosshair_vline.get_xdata()[0]
     current_state.bin_vs_x = True
     current_state.bin_vs_y = True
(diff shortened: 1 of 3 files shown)


docs/artists.rst
```diff
@@ -98,12 +98,21 @@ datasets.
...
+
+================  =======================================================================
+key               action                                                       
+----------------  -----------------------------------------------------------------------
+tab / ctrl+tab    cycle focus between the sliders and the plot 
+left/right arrow  decrement/increment slice (slider focus) or change y slice (plot focus)
+up/down arrow     change x slice (plot focus)
+================  ========================================================================
+
+Note that the up/down arrow navigation overrides the built-in undo/redo action of the qt viewer.
```
ksunden: left/right is the built in undo/redo.

May also wish to not that `c`/`v` and the top toolbar gui buttons retain the undo/redo action

WrightTools/artists/_interact.py
```diff
@@ -14,6 +14,32 @@
 __all__ = ["interact2D"]
 
 
+class Focus:
```
ksunden: Strange use of a class, to be honest.
Something that only implements `__call__` should probably just be a function plus a private variable, especially when the business logic in `__init__` is the same as in `__call__`

That said, I don't think it warrents delaying merging to fix this up, it is not in `__all__` and is thus a private implementation detail.

ksunden: Could also be implemented as a method of `Bunch`, as that is how it is used internally, maybe... Bunch is itself a weird object because it inherits from dict and sets `__dict__ = self`

ksunden:
Approved, feel free to merge upon CI passing

PR #908, opened 2019-05-30
Large file parsing in PyCMDS

ksunden:
This is still relatively rudimentary.

The first question is this: Do we wish to have this in WT permanently? How useful is it, should it just be a script distributed with the abnormally large datasets/as part of such a publication, or should it be always accessible.

If the former, then the file as is should do fine, I think.
If it is the latter, my current leaning is to not have this be a new file, but instead a new function in `_pycmds.py` with shared code extracted into private functions. and the parsing step itself being the only difference.

Thoughts and comments welcome.


WrightTools/data/_pycmds_large_file.py
```diff
@@ -0,0 +1,196 @@
...
+                    sh[j] = 1
+            print(f"{ax['name']}_centers")
+            data.create_variable(name=ax["name"] + "_centers", values=centers.reshape(sh))
+    # get assorted remaining things
+    # variables and channels
+    try:
+        signed = iter(headers["channel signed"])
+    except KeyError:
+        signed = itertools.repeat(False)
+    for index, (kind, name) in enumerate(zip(headers["kind"], headers["name"])):
```
ksunden: Only superficial differences to the usual `from_PyCMDS` occur before this point: namely this for loop uses enumerate instead of range and the file is not closed/a full 2D array read. Those lines can be moved to after all headers are parsed, they were placed before so the file could be closed earlier, releiving a small amount of memory, but the effect should be minimal

In this version, this loop only creates variables and channels, does not fill them, filling is done afterwards.
In the original, this loop performs both tasks.
This version also assumes float64 for all fields, the usual version will do whatever genfromtxt gets originally.

WrightTools/data/_pycmds_large_file.py
```diff
@@ -0,0 +1,196 @@
...
+                continue
+            if name == "time":
+                name = "labtime"
+            if "wa" not in headers["name"] or name in ("wa", "array", "array_signal"):
+                h5py.Group.__getitem__(data, name)[index + (...,)] = arr[:, i]
+            else:
+                h5py.Group.__getitem__(data, name)[index + (...,)] = arr[0, i]
+        arr = np.genfromtxt(file_, max_rows=frame_size)
+    file_.close()
+    # axes
```
ksunden: Nothing changed after this

untzag:
@ksunden if we distribute this, would it be practical to test against it?

in a perfect world, I agree this should be distributed within WrightTools

ksunden:
It works fine on the `w1_wa` dataset we already distribute
(In fact it should read in any dataset we throw at it)

darienmorrow:
@ksunden Is this PR still a draft? Do you want to include some rudimentary tests as recommended by @untzag?

IMHO we should have this method subsumed by `from_PyCMDS` and just accessed by a kwarg (`memory_safe`, or `large`, or something of the sort).

PR #909, opened 2019-06-03
BUG: rework tolerance for axes, specifically with centers

ksunden:
- [x] Add test


ksunden:
ping @untzag

untzag:
@ksunden can you write some sentences to describe what exactly this fixes, and how?

ksunden:
Essentially, there was an edge case (which I only ever saw come into play when there were two `centers` arrays) where the `tolerance` was not set at all (thereby incorrectly using the tolerance stored in the data file for axes which should have a tolerance of 0)

This reworks that a bit to actually set in all cases it should be set to either 0 or infinity.

The other thing this does is remove the `assert`, which are not run in all instances (`python -O` disables asserts) (We have a handful of others that I think we may wish to remove though all others are `assert self.x is not None` for caching of `@properties`)

ksunden:
The new test data file had read in incorrectly, specifically `w1_Delay_1` read in with a shape of (1, 1, 11), instead of (25,1,11), collapsing down an axis that it should not have.

ksunden:
see also wright-group/attune#58 for another example of a dataset which had been incorrectly parsed

PR #915, opened 2019-08-12
Travis deployment automation

ksunden:
Whitespace changed because travis' cli util apparently likes 2 space indents in yaml, and automatically adds secure bits.


ksunden:
we shall test this with the next release (slow iteration if there's anything wrong, I suppose, but should be good, I think)

PR #931, opened 2020-01-13
MAINT: Units: Prefer numexpr over eval

### LINKED ISSUE #920: consider using NumExpr as alternative to eval
Within the units library, we use eval.

https://github.com/wright-group/WrightTools/blob/master/WrightTools/units.py#L96-L135

We could use NumExpr... our usage case seems to be exactly what the package was made for.

https://numexpr.readthedocs.io/en/latest/user_guide.html

ksunden:
Closes #920

### DIFF
--- a/WrightTools/units.py
+++ b/WrightTools/units.py
@@ -5,9 +5,10 @@
 
 
 import collections
+import warnings
 
 import numpy as np
-import warnings
+import numexpr
 
 
 # --- define --------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


WrightTools/units.py
```diff
@@ -20,7 +21,7 @@
 angle = {"rad": ["x", "x", r"rad"], "deg": ["x/57.2958", "57.2958*x", r"deg"]}
 
 # delay units (native: fs)
-fs_per_mm = 3336.
+fs_per_mm = 3336.0
```
ksunden: This needs to be added to the numexpr locals, I think (at least wish to be explicit)

ksunden: In point of fact, it works, so just going to leave it for now, but being more explicit may be nice.

PR #932, opened 2020-01-13
MAINT: Fix mpl deprecations

### LINKED ISSUE #890: Matplotlib deprecation for clim

ksunden:
Closes #890

### DIFF
--- a/WrightTools/artists/_helpers.py
+++ b/WrightTools/artists/_helpers.py
@@ -978,6 +978,8 @@ def set_fig_labels(
         if ax.is_sideplot:
             continue
         try:
+            # [row|col]span were introduced in matplotlib 3.2
+            # this try/except can be removed when supprot for mpl < 3.2 is dropped
             rowNum = ax.get_subplotspec().rowspan.start
             colNum = ax.get_subplotspec().colspan.start
         except AttributeError:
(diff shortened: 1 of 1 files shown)


ksunden:
This is two changes, one for `set_clim`, which was deprecated in 3.1, removed on current master mpl, it has been a no-op that the only complaint recieved was the warning message.

The other is for `[row|col]Num`, which was changed by matplotlib/matplotlib#13544

There may yet be a more elegant way to loop through precisely the axes we want, looking at that, but this quick fix should work for now, and does not (yet) remove support for any currently supported mpl

Note:  in looking at that diff, I noticed a `label_outer` function that may be useful for `set_fig_labels` (though it does not let you hand it the text, just ensures visibility of axis labels)

PR #934, opened 2020-01-22
MAINT: refactor savefig to accept mpl kwargs

ksunden:
Maintains the same defaults, but allows user to override (and does not require adding kwargs manually)


ksunden:
ping @darienmorrow

ksunden:
Test failure is not related, about to open a PR to address, osf changed their URL from accepting either of two formats to only one

PR #936, opened 2020-01-22
DEP: Deprecate pcolor_helper, remove from internal plotting functions

ksunden:
Depends on matplotlib/matplotlib#16258.

The PR on mpl introduces a new kwarg that eliminates the need for pcolor_helper.
Once merged and released, we should bump our min supported mpl, and deprecate pcolor_helper.

For the time being, the function just has a warning, but is unchanged, so if callers call it directly it will work.

The mpl implementation yeilds subtly different default x/ylim values, which such that the whole pixel of boundary data is visible. Our `pcolor_helper` wound up with the x/ylim being the data point itself.
Plots are identical (modulo perhaps some single pixel edge effects) if xlim and ylim are set manually.
I personally actually prefer the way mpl did it.

This PR Requires a bump to the mpl version in `setup.py` as well as filling out the proper version numbers in the docstring of pcolor_helper.

For now, we have not declared a concrete deprecation period/policy, so I'm not declaring when it will be removed (heck, we may just say "no new code should use it", but leave it there for legacy code indefinitely). That said, I'm in favor of cleaning up our workarounds as mpl introduces proper fixes (or we discover the intended way to do things that we hacked and slashed).

As an aside, this will also _decrease_ test coverage, I believe, as we do not have a targetted unit test at pcolor_helper (it was being run previously, but only by the internal calls). I'm inclined to say this doesn't matter, as it is deprecated.
We do, however, get rid of the circular import problem.


ksunden:
The above PR has been merged into MPL, and will be included in the next feature release (should be 3.3.0, I think, incidentally, the same as our version number likely to include this change)

ksunden:
Closing and reopening to get CI to run... MPL 3.3 is out now

ksunden:
Test failure on push is already fixed by #935, could be fixed by rebasing if desired, but really unrelated to this PR

PR #940, opened 2020-03-17
add kwarg passing to artists.corner_text

darienmorrow:


WrightTools/artists/_helpers.py
```diff
@@ -216,7 +222,7 @@ def corner_text(
     # apply text
     props = dict(boxstyle="square", facecolor="white", alpha=background_alpha, edgecolor=edgecolor)
     args = [v_scaled, h_scaled, text]
-    kwargs = {}
+    kwargs = {**kwargs}
```
ksunden: This is a no-op that can just be deleted

ksunden: I also noticed we would overwrite some kwargs if they were passed, so I just made that change, including removing this line

Issue #809, opened 2018-11-28
gallery example of moment calculation, plotting


untzag:
see #786

Issue #804, opened 2018-11-28
sideplot gallery example


darienmorrow:
e.g. Figure 3 of Czech2015

Issue #499, opened 2018-02-25
animation example


untzag:
animation example on wright.tools

ksunden:
https://github.com/sphinx-gallery/sphinx-gallery/issues/150

This may be a bit more of a pain than I initially anticipated

ksunden:
Requires a toolchain update to sphinx-gallery 0.2.0. We are currently pinning the version because of some clashes that previous versions had. Will have to investigate further...

ksunden:
Currently we fail to build under 0.2.0, due to it for some reason being unable to find the datasets. I think it may have to do with pickling, actually... which is frustrating in and of itself...

darienmorrow:
Is this ready for a PR?

ksunden:
I do not know... probably...
@emkaufman had an example written (in the `animation` branch) that can be rebased on top of master to get a starting point...

For it to render well, probably also involves writing a "scraper" to get the GIF output:

https://sphinx-gallery.github.io/stable/advanced.html#example-2-detecting-image-files-on-disk

(Taking the PNG version there and replacing png with gif)

PR #941, opened 2020-03-19
add moment and sideplot gallery example. Closes (#809, #804)

darienmorrow:


darienmorrow:
@ksunden can you check on the CI stuff? They passed the first time I pushed, but failed the second time.

ksunden:
Test failure is real... It's the same problem we've seen in the past, which we solved largely by reordering tests until they worked... which was annoying, and never truly figured out why our attributes in `datasets` disappeared...

darienmorrow:
@ksunden So what is to be done in this case?

ksunden:
@darienmorrow I honestly have no idea... I think perhaps the best course of action is to go back to a constructed data object.

The bug is, I think, in sphinx (though may be sphinx-gallery), and I have no clue why it comes up, where attributes disappear it seems after they have been used by the doc building...

untzag:
@ksunden do you think there is any good to be found in contacting the sphinx-gallery people and asking them for their advice?

ksunden:
@untzag I honestly doubt it...

ksunden:
It's really a strange condition... And I don't know what's causing it... we use the same dataset in another example (label_delay_space)... And we use a different PyCMDS dataset in two examples (so it's not due to using it twice, at least not as a sufficient condition).

If we could come up with a more minimal example, I think then maybe we could do well to reach out to upstream projects, but as it is, I don't think this is an informative example... the datasets package is populated in a strange way, theres just too much going on for my taste in terms of being able to tract what is happening

untzag:
@ksunden well that makes me very frustrated...

are there ANY ideas we have to permanently remove this thorn? up to and including going for homebuilt + pdoc for WrightTools.

would it help us if we broke the datasets subpackage into a separately distributed package that was designed differently? this problem MUST come from how sphinx caches WrightTools, right?

know that you don't actually understand the problem @ksunden, so not trying to put the work of figuring it out onto your shoulders... still would appreciate your thoughts

ksunden:
Any number of those options would be worth investigating, in my opinion...

One (relatively) easy check I'm considering trying out  is reworking the datasets package to use `__getattr__` dynamically, rather than using `setattr`... I don't particularly like that path because it would mean that things like tab completion when using these datasets in an interactive setting would not work (and these are not particularly easy to remember full names, either...

I would not be opposed to some pdoc and/or jinja template solution, but that is a fair amount of work (yet may be worth it)
An intermediate step would maybe be to make our own version of only the gallery and not use sphinx-gallery... Still not sure that's what I want, but...

Here is the code that is supposed to reset WT for sphinx gallery...
```
def reset_wt(gallery_conf, fname):
    # This is an awful hack because python does not allow unloading modules
    # This is, however, the same thing upstream sphinx-gallery does for
    # seaborn, so it's not _so_ bad I guess.  2019-04-07 KFS
    for module in list(sys.modules.keys()):
        if module.startswith("WrightTools.datasets"):
            del sys.modules[module]


sphinx_gallery_conf = {
    "examples_dirs": "../examples",
    "filename_pattern": "/*.py",
    "gallery_dirs": "auto_examples",
    "download_section_examples": False,
    "backreferences_dir": os.path.join("gen_modules", "backreferences"),
    "reset_modules": ["matplotlib", reset_wt],
}
```

ksunden:
https://sphinx-gallery.github.io/stable/configuration.html#resetting-modules

The resetting is documented as being on a per file basis... Which made sense... until now for... some reason...

darienmorrow:
@ksunden I am a little lost on what the actual issue is and how to fix it. Is there anything actionable I can do here? Should I make a "copy" of the dataset I use into the .wt5 format? If there is something actionable, should the same action be applied to #499?

ksunden:
@darienmorrow I have _no idea_ what the problem is... 

Any of the solutions proposed by blaise or yourself _might_ work, I simply do not know.

For my part, I am 100% fine with not using real data for this demonstration, which was shown to work.

darienmorrow:
Ok. I have switched to using simulated data. Below us the output. @ksunden can you review? Hopefully CI checks pass this time.

image

ksunden:
Travis still having trouble communicating that it passed, merging with admin rights

PR #948, opened 2020-08-26
Correct split units for output data objects and print

ksunden:
## Changes

Split did not correctly reassign units to the axes to be the same as the input data object. additionally, it printed the incorrect units in verbose mode

## Checklist

- [x] updated CHANGELOG.md
- [x] tests pass


kameyer226:
Assert for shapes during split tests works,  however, verbose output is incorrect.   See for example `test_split_multiple`.

Response at terminal for this example:
`split data into 3 pieces along <wm>:
  0 : -inf to 19705.00 wn (15, 1, 1)
  1 : 19705.00 to 20605.00 wn (18, 1, 1)
  2 : 20605.00 to inf wn (2, 1, 1)`

(should be (15,11,11) etc )

ksunden:
That is showing the shape of the split expression, not the overall data object, it is correct (in that sense), and arguably more useful as it makes it really clear what the data looks like on the split expression

ksunden:
All of the changes to things outside of split were just due to changes in the autoformatter we use, no actual code changes, just whitespace

ksunden:
Actually... I looked again, not sure when I made those changes... they seem to be _correct_, and the tests still pass, so not going to undo them... but I had a dirty repo when I started (my fault) and didn't carefully look at the diff (I skimmed over it and thought "oh, hey, black did that, not sure why")

I did also have some playing with a black and white logo in there... revert that file in #951

kameyer226:
Why not have both?  There are reasons, like you just mentioned

ksunden:
We could do both, but I was just playing around, wanting something to 3D print, it was never intended for use... (and overwrote the color version)

For now, not worth spending time on making it produce both, in my opinion (it's not hard, but requires writing new lines nonetheless)

PR #953, opened 2020-09-02
More precise fs per mm

ksunden:
Might as well be more precise...

## Checklist

- [x] updated CHANGELOG.md
- [ ] tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 ### Changed
 - misc text effects within `artists.corner_text`
 - deprecate `pcolor_helper`, remove from internal plotting functions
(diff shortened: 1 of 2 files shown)


kameyer226:
Careful,  the speed of light in AIR is NOT 299,792,458 m/s.   May not want to use the air value (closer to 299,702,547 but we all know about variances).

kameyer226:
One could calculate index of refraction of various gases/humidities used inside the apparatus at the delay area at Lab Temp/Pressure and truncate at the point where the differences between them diverge.   Even that makes assumptions for future work but it is probably better than the current number we use.   (The number should be around 3,336.6  fs/mm, not 3335.6, for our context of this number)

ksunden:
Closing as we simply don't have 5 sig figs anywhere anyway, so no sure it really matters

PR #957, opened 2020-09-09
using pint for units

### LINKED ISSUE #39: compatability with matplotlib.units
I have just recently discovered that matplotlib has a units API. Making WrightTools objects compatible with this api may streamline plotting outside of `wt.artists`.

http://matplotlib.org/api/units_api.html

### LINKED ISSUE #132: units
I'm opening this issue because I want to be 100% open and honest about WrightTools and units. I know that units support is very important to WrightTools, and I want the package to be as good as possible. I also know that there is much more that WrightTools could do with units, and I'm open to the idea of overhauling units support moving forward.

That said, units get really hard really quickly. Our current humble home-brew strategy is simple and foolproof, which goes a long way. I'd prefer to not change the units API, if possible.

Should channels have units?

Should we fall back on a units library? Such as...
1) [pint](https://pint.readthedocs.io/en/0.8.1/)
2) [natu](https://kdavies4.github.io/natu/)
3) [magnitude](https://github.com/juanre/magnitude)
4) [scimath](https://github.com/enthought/scimath)
5) [quantities](https://pythonhosted.org/quantities/)
6) [units](https://bitbucket.org/adonohue/units/)
7) [SymPy](http://docs.sympy.org/latest/modules/physics/units/index.html)
8) [astropy](http://docs.astropy.org/en/stable/units/)
9) more?

### LINKED ISSUE #449: Use a 'native' energy unit that is proportional to energy
Currently we use `nm` as the "native" energy unit.
I feel that it may be cleaner to have the native unit be one of the other units, I propose `wn`.
This would allow simpler proportionality constants for most unit conversions (other than `nm`) rather than inverse relationships.

This would also allow for some subsystems to reliably do additions in the native unit for energy. (This is not implemented, nor do I think we have direct plans to do so, but this change would be necessary in the event we wanted to)

ksunden:
## Changes
This is a proposed major overhaul to the units system, using Pint as the backend rather than rolling our own.

We have seen several instances of our conversions being incorrect (e.g. #955).

At this stage this PR represents a minimum viable product

I tried to retain backwards compatibility where possible.

Still to do:

- [x] lists of compatible units... prefixes make this hard, also just may wish to have a simple list of common, not make that exhaustive, since pint really gets to some weird ones
- [x] mm_delay behavior... this is best handled in pint with a context, rather than a new unit definition, the unit can be defined for backwards compat
- [x] Update docs, completely out of date, but waiting to put effort there until I receive some feedback

## Checklist

- [x] updated CHANGELOG.md
- [x] tests pass 
 
Closes #132 
Closes #39
Closes #449


tests/data/axis/convert_axis.py
```diff
@@ -43,8 +43,8 @@ def test_w1_wa():
     assert data.wa.units == "wn"
     data.wa.convert("eV")
     assert data.wa.units == "eV"
-    assert np.isclose(data.wa.max(), 1.5802564757220569)
-    assert np.isclose(data.wa.min(), 0.6726385958618104)
+    assert np.isclose(data.wa.max(), 1.5800551001941774)
+    assert np.isclose(data.wa.min(), 0.6725528801867734)
```
ksunden, line 47: These constants were computed _less_ precisely with our own units library.

The same dataset (and thus constants checked) was used when converting axes, data object, and dataset directly)

WrightTools/units.py
```diff
@@ -152,28 +91,27 @@ def get_symbol(units) -> str:
...
-        return r"T"
+    if dimensionality == "[temperature]":
+        return "T"
     else:
         return kind(units)
 
 
 def get_valid_conversions(units) -> tuple:
+    return ()
     try:
```
ksunden, line 115: Yes, this introduces unreachable code...

ksunden:
If we committed to this, there are about 10 places in PyCMDS that would need updating, primarily where we use `kind` to get a list of available units, or as a proxy for "Can convert to"...  though the latter gets a little murky because length can be both "delay" and "wavelength", though I think we only ever use time as delay, so maybe that resolves it...

ksunden:
Revisiting now, so far all I've done is rebased on current master

ksunden:
Okay, tests are passing locally

for backwards compatibility/expected behavior when you do things like `data.convert("ps")`, I have put in a list of "blessed units".

blessed units are those that prior wt handled (minus a few rarely used ones that don't have conversions anyway like uJ/cm^2)

basically if you ask for what conversions _can_ happen it will draw from this list (rather than trying to do an exhaustive list of units in the registry, which would be easy, but not useful)

Additionally if you are trying to convert between two things with in the blessed list, old "energy" and "delay" groups are retained. If either unit is not in the blessed list, it falls back to "does pint know how to convert between these two".

This could still lead to some strange behavior, but only if you veer away from the "blessed" units (and we can expand them if we wish) for example, if you were to convert a data object to "fortnight", _all_ energy and delay axes(/variables) would be converted, since pint knows how to convert all of the spectral units into time (via the speed of light and the spectral context)

I certainly think the docs _need_ to be updated prior to merge, as well as probably several additional tests (suggestions welcome)

But all external behavior touched by tests is consistent.

ksunden:
codeQL failure is a real problem that should be addressed (I'm pretty sure it is an indentation problem and that the block that computes the variable in question should be dedented so it always executes)... but not relevant to the change at hand, so ignoring for this PR

untzag:
I suggest `blessed_units` lists for each kind on the top level. This will be a good bridge for tools like yaqc_cmds that expect to be able to get all units of a given kind.

Speaking of kind, we should keep the interface to kind the same for external tools. We can raise a deprecation warning. If we get non-blessed units kind can be `None`.

untzag:
@ksunden this looks largely OK to me. However this is a big change for the library. I think we should get approval from multiple maintainers before merging.

darienmorrow:
@ksunden Here are a few things I think would be required for merging:
- inline comment about purpose of `blessed_units` tuple
- wright.tools documentation about units
- Appropriate version kick. I am thinking `3.4`
- Proper representation of `wn` unit. image is incorrect. It should be something like: _$\left(cm^{-1}\right)$_.
- Proper naming representation of axes when using nonstandard spectroscopy units. (See example below). When I call `d.field.label` I expect to get `'$\\mathsf{field\\,\\left(T\\right)}$'` and not what I currently get on this branch: `'$\\mathsf{[mass] / [current] / [time] ** 2\\,\\left(T\\right)}$'`.

```
import WrightTools as wt
import numpy as np


field = np.linspace(-1,1,31)[:,None]
acc = np.linspace(-1,1,31)[None,:]
chan = field * acc 
d = wt.Data(name='test')
d.create_variable('field', units='T',  values=field)
d.create_variable('acceleration', units='m/s',  values=acc)
d.create_channel('arr', values=chan, units = 'T', signed=True)
d.transform('field', 'acceleration')
wt.artists.quick2D(d)
```
image

untzag:
@ksunden let's add some non-blessed unit examples to the gallery so we can be sure that our labels are doing something sane

ksunden:
@kameyer226 the "blessed" unit list is simply those that we had support for in the past (not that we actually used some of them...)

We typically do not put units on DAQ readings, since the unit of V is pretty meaningless (we aren't going to be doing conversions, usually normalize it anyway.

The use of pint as a backend is in part to allow these units when it makes sense, though. Just because it is not "blessed" doesn't mean it can't be used. Whereas in prior state we had _no_ support for units of power/current/etc.

ksunden:
@darienmorrow The behavior of adding the dimension to the label is mostly just an attempt at getting something semi sensible in place of the symbol we used to use... admittedly it is quite ugly for compound dimensionality... 

We have never had it so that the name of the field appears in the label (other than subscripted, which would still happen here if you set the "label" field on the axis)

We can change that behavior if we want... that was mostly standin behavior

re `wn`, I need to check back in with pint units to hopefully figure that one out... otherwise need to special case that one...

bumping version will not be part of this PR, but I agree that it should be done prior to release

ksunden:
@untzag I have no interest in maintaining the `kind` dictionaries.... they are not used at all internal to WT, and it is pretty trivial to get rid of their use in yaqc-cmds.

I do not really consider them truly public (even though we did not put the _ in the name... I never thought it was good to rely on that, and don't think we do outside of yaqc-cmds)

the impedence mismatch between "kind" and pint is just annoying here...

untzag:
I do not agree that it is trivial to get rid of this behavior in yaqc-cmds. How are we going to make graphical interfaces with sane (limited) choices for unit conversions without maintaining sets of kinds somewhere?

If your proposal is to maintain this set within each graphical interface that's fine... is that your proposal?

ksunden:
That is precisely what the `get_valid_conversions` does....

it uses the `blessed_units` list by default (though any list can be passed)

I think it is not useful at all...

untzag:
Please?

WrightTools/units.py
```diff
@@ -152,43 +129,43 @@ def get_symbol(units) -> str:
...
+    if dimensionality == "[mass] / [time] ** 2":
         return r"\mathcal{F}"
-    elif kind(units) == "pulse_width":
-        return r"\sigma"
-    elif kind(units) == "temperature":
-        return r"T"
+    if dimensionality == "[temperature]":
+        return "T"
     else:
         return kind(units)
```
untzag, line 148:
```suggestion
        return units
```

untzag: actually I changed my mind

untzag, line 148:
```suggestion
        return "{:L}".format(ureg.Unit(units))
```

untzag (approved):

WrightTools/units.py
untzag: https://pint.readthedocs.io/en/0.7.2/tutorial.html#string-formatting

ksunden:
A poor implementation is worse than no implementation, and I do not wish to put the effort into making a good implementation.

It will take all of 10 mins to convert yaqc-cmds, pretty sure

WrightTools/units.py
ksunden: I do not think this is what we want...

That will result in (in Dariens example of Tesla) `tesla (T)`

I think perhaps the right call is to return None, and then place the `axis.name` in the string in the axis class...

Additionally, pint's Latex formatting uses `mathrm`, where we have historically prefered `mathsf`

ksunden:
Figure_1


@darienmorrow here is that same example with the most recent commit which addresses the fallback behavior if no symbol is found.

Arguably `\lambda` should not be used for _all_ length measures... hard to know where the dividing line is though (only nm, nm and micron? what about smaller prefixes?)

WrightTools/units.py
```diff
@@ -152,48 +138,50 @@ def get_symbol(units) -> str:
...
-        return r"\sigma"
-    elif kind(units) == "temperature":
-        return r"T"
+    elif quantity.check("[temperature]"):
+        return "T"
+    elif ureg[units] in (ureg.deg, ureg.radian):
+        return r"\omega"
     else:
-        return kind(units)
+        return None
```
ksunden, line 159: this method may be cleaner as a dict... (of dimensionality -> str, with angle/other dimensionless units handled separately)

For _right now_ I'm inclined to keep it in the if/elif structure until and unless we decide to add more to it.
But I'm open to suggestions of what to add, and if given will likely refactor.

untzag: I like leaving it as an if elif tree

ksunden:
I believe this is ready for merge now, willing to add to docs if requested, though.

docs/units.rst
```diff
@@ -9,54 +9,57 @@ You can use it directly, if you wish.
...
+   >>> wt.units.convert(2., 'eV', 'nm')
+   619.9209921660013
+
+Under the hood, the WrightTools units system is backed by a `Pint <https://pint.readthedocs.io` unit registry.
+This registry has all of the units in the default Pint registry, plus a few added by WrightTools (including some extra aliases of pint units for backwards compatibility with older WrightTools).
+Additionally this unit registry automatically enables the spectroscopy context as well as a custom context for handling delay behavior (which inserts a factor of 2 as delays typically increase the path of light by two times their physical movement).
+
+The pint unit registry can be directly accessed at ``wt.units.ureg``.
+
+To query wheteher a specific conversion is applicable, you can use :meth:`wt.units.is_valid_conversion`.
```
ddkohler, line 21: whether

untzag (approved):

untzag (approved):

PR #962, opened 2020-10-02
Gaurd against trying to write readonly files

ksunden:
## Changes

Current behavior will _always_ try to write to attrs at init time, this PR prevents such writes where unecessary, allowing for read only files to  be opened.

What this does _not_ do: 
- provide an in house way of opening in read only mode (e.g. param/kwargs pass through to `wt.open`)
   - Only way to do it is by first opening the file in read only mode using h5py

## Checklist

- [x] updated CHANGELOG.md
- [x] tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
-
 ### Fixed
 - representation of units with micro prefix
+- Do not attempt to write attrs when file is read only
 
 ## [3.3.0]
 
--- a/WrightTools/_group.py
+++ b/WrightTools/_group.py
@@ -66,9 +66,11 @@ class Group(h5py.Group, metaclass=MetaClass):
         h5py.Group.__init__(self, bind=file[str(path)].id)
         self.__n = 0
         self.fid = self.file.id
-        self.natural_name = name
+        if "name" not in self.attrs.keys():
+            self.natural_name = name
         # attrs
-        self.attrs["class"] = self.class_name
+        if "class" not in self.attrs.keys():
+            self.attrs["class"] = self.class_name
         if "created" not in self.attrs.keys():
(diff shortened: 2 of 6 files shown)


kameyer226:
love to know why the checks failed.   Do you want to throw an exception if readonly file found?

kameyer226 (dismissed):
Is the guard so that it doesn't throw an exception?  My comment would be invalid then.

ksunden:
I do not know for sure why the check failed.

We write to the file when it is opened (even though we are often just writing the same contents)

This caused it to raise an exception when passed a read only file (we found this in pycmds land, when doing some things with Single Writer Multiple Reader mode)

This just detects when it is in readonly mode (or SWMR mode, which is why some checks include None, since for some reason h5py sets the file mode to None in this mode), and prevents the things that write attributes (which if you are reading an extant file in read only mode must already be filled out and will not change, so no harm in not doing so... the only thing it does is prevent some of our cached values from populating, which is minor consideration)

The single check that failed has to do with behavior on copy, which indicates that maybe one of the changes prevents something from being properly updated in that scenario, and should have the condition changed, I suspect the ones that are just asserting that the key is there, rather than checking the file mode, but I will take a closer look prior to merging

ksunden:
I think I figured out what is happening, requires a little bit more targeted surgery to get around.

Boils down to these lines:
```
if "name" not in self.attrs.keys():
     self.natural_name = name
...
if parent.name == self.name:
            pass  # at root, dont add to item_names
elif self.natural_name.encode() not in parent.attrs["item_names"]:
            parent.attrs["item_names"] = np.append(
                parent.attrs["item_names"], self.natural_name.encode()
            )
```

Basically, on copy, the name _changes_, so it is already in the attrs dict, but is supposed to be updated.
The code here prevents it updating.

kameyer226:
I'm not completely zeroed in, but would you have to append the new item_name , then delete the old one from the dict afterwards?

ksunden:
Appveyor failure appears unrelated, rerunning

untzag (approved):

Issue #960, opened 2020-09-25
Data.copy() does not inherit units of individual axes


ddkohler:
```
d1 = wt.Data()
d1.create_variable(name="color", values=np.array([1]), units="nm")
d1.create_variable(name="length", values=np.array([1]), units="nm")
d1.transform("color", "length")

d1.color.convert("wn")
d2 = d1.copy()
print(d1.units, d2.units)  # d1 has units of (wn, nm) , d2 has units of (nm, nm)
```

This is problematic e.g. for controlling the axes units for `interact2D`, which copies the data object for plotting.

darienmorrow:
This is a "gotcha" I have often encountered. 

`d1.color.convert("wn", convert_variables=True)` does what you intuitively want. 

I'm not sure what default behavior should be.

darienmorrow:
Oops, upon testing, I don't think I am correct.

ksunden:
This boils down to the attrs dictionary not being updated in `axes.convert`. this could be fixed by calling `self.parent._on_axes_updated()` in that method.

It can be worked around by calling `transform` _after_ the `convert` call

ksunden:
(as `transform` and `Data.convert` both call `_on_axes_updated`)

ksunden:
closed by #974

PR #974, opened 2020-10-28
Data copy individual axis units

ddkohler:
## Changes

Resolve issue #960 

## Checklist

- [x] updated documentation (N/A) 
- [x] updated CHANGELOG.md
- [x] tests pass

### DIFF
--- a/tests/data/axis/convert_axis.py
+++ b/tests/data/axis/convert_axis.py
@@ -23,7 +23,7 @@ def test_convert_variables():
     data.close()
 
 
-def test_convert_variable():
+def test_units_preserved_on_copy():
     d1 = wt.Data()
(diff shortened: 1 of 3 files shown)


tests/data/axis/convert_axis.py
```diff
@@ -23,6 +23,20 @@ def test_convert_variables():
     data.close()
 
 
+def test_convert_variable():
```
ksunden, line 26: I would like a more descriptive name for this test, it is too similar to the test above it (in name), which could cause confusion if/when one of them fails (especially if the other passes)

ddkohler: Changed to more descriptive name.

ksunden (approved):

Issue #757, opened 2018-10-11
consider dropping appveyor


untzag:
https://blog.travis-ci.com/2018-10-11-windows-early-release

ksunden:
Another option is Azure DevOps, Numpy as added them
Spreading out a bit can be nice because sometimes they limit concurrent runs, so having different services will finish faster than having them all the same.

It is true that it is more difficult to maintain the config files, but that cost is relatively minimal, all things considered.

Current leaning: Appveyor is not a bottleneck (though we do only test one python version to ensure that), so "if it ain't broke, don't fix it", but open to discussion

untzag:
not a big enough deal to maintain the issue

PR #976, opened 2020-11-03
Switch from travis/lgtm to github actions

ksunden:
## Changes

Switching to github actions in the wake of the travis changes for OSS (and further wake of the transfer of ownership of travis)

This incurs no code changes

 Edit:

apparently the new h5py string handling breaks us... pinning version to under 3 for right now to get (hopefully) green CI... but I intend to do a proper fix either pinning _to_ v3 or inclusive of both v2 and v3 prior to release, so not doing any additional things (e.g. changelog entry)


ksunden:
Py 3.9 takes _forever_ to build scipy in gh actions (not sure why...) the 3.7 and 3.8 tests pass fully before the deps are built

It is only a problem until scipy release wheels on 3.9

My leaning is to remove 3.9 from the test matrix for the time being (until scipy release 3.9 wheels)

Alternative would be to figure out dep caching, as it only has to be done once then

ksunden:
My intent is to move off of appveyor as well (#757), but want to complete the travis transition first, so that will be a separate PR to do windows via actions

ksunden:
Hmmm... weird thing is, after the 30 mins it took to run the 3.9 tests, a single test failed, namely a place we were expecting to raise `ValueError` instead raised `RuntimeError`... quick survey of the h5py source did not indicate why that might be, and I'm inclined to ignore it still, but keep an eye out for it in the future

The fact of the matter is that _which_ exception it raises there (when trying to access a closed file) is less important than _that_ it raises an exception...

untzag (approved):

PR #977, opened 2020-11-04
Fix string handling in h5py 3.0

ksunden:
## Changes
H5py 3.0 caused some small changes that were caught by our previous tests (namely calling "decode" on strings)

## Checklist

- [x] updated CHANGELOG.md
- [x] tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 - Return min and max when file is read only
 - try/except the os.close call for tempfile (may already be gone in SWMR mode)
 - data objects retain units of individual axes when copied
+- h5py 3.0 string handling
 
 ## [3.3.1]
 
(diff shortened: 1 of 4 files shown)


untzag:
shall we pin to >3.0?

ksunden:
Hmmm... tests do pass locally... 

Precommit is just black not being the right definition to get it installed properly

ksunden:
I don't see any particular reason to pin to h5py 3... the change should support both (as it is only taking _out_ one line of code if it is not needed, no change if it is needed, and it continues to be needed to read older hdf files...)

ksunden:
something weird being fixed by my most recent commits... locally tests pass fine, and it raises valueerror as it always has, but on CI its raisin runtimeerror instead... not sure why...

ksunden:
Ignore pre commit, I will fix it in a separate PR

ksunden:
The reason for the different errors is that my local one is built off of debian provided hdf 1.10.6, whereas CI is using wheels built off of 1.12.0.

After consultation with tacaswell, it seems like they _intend_ to raise the more specific errors that are historically raised, but the hdf group doesn't consider codes used to be stable, so their mappings are out of date with the most recent hdf version.

For our purposes, I'm fine with including runtimeerror, I guess... but may follow up with a proper issue filed with h5py detailing what we observed so that hopefully it can be more common for others.

untzag (approved):

PR #981, opened 2020-11-06
Add py3.9 tests

ksunden:
Scipy, numpy, h5py now all have 3.9 wheels (mpl does not yet, so we will see)


ksunden:
If 3.9 tests take longer than appveyor still, we should hold off

ksunden:
Superceded by changes in #982 (just included this there, as I was changing the workflow file more significantly anyway)

PR #983, opened 2020-11-16
Add some more try/except around things that can fail when read only

ksunden:
…but aren't really necessary in those cases anyway

## Changes

## Checklist

- [x] updated CHANGELOG.md
- [x] tests pass

I'd like to add a test, this in particular is for cases in which _one_ of min/max is cached, and then you ask for the other one in readonly mode...


untzag (approved):

ksunden:
rerunning ci because of one failure which I think is unrelated (it failed in build_docs, hard crash on Win py3.6, and only that one, I think it is a random, rare failure, not a real problem, but checking anyway)

PR #996, opened 2021-03-13
Convert verbose

darienmorrow:
## Changes

Propagates verbose kwarg to convert for data.split

## Checklist

- [x] updated CHANGELOG.md
- [x] tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 - Collection.convert method for converting all units of a single kind of a collection's data to another unit (with test script)
 
 ### Fixed
-- Absence of verbose kwarg propogation in data.split
+- Absence of verbose kwarg propogation in data.split, silenced convert call.
 - Further handling for readonly files
 - Improved chopping with axes that span the kept axes removed
 - Timezone offset for `wt.kit.TimeStamp.RFC3339`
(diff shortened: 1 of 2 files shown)


ksunden:
In this instance, I may actually lean towards calling convert with `verbose=False` it is an implementation detail, not something the user really ever needs to know... in fact line 1756 converts it back so it is only converted during computation (and that line needs the same kwarg passed regardless)

darienmorrow:
@ksunden I could be wrong: I recall that axes, constants, etc _do not_ have `verbose` exposed. So it should not be passed to them. The reason it is set to `True` is that we generally have `verbose` set to True throughout the entire stack. I would actually prefer if `verbose=False` was the norm in the stack.

ksunden:
Okay... arguably the axis/constant/variable `convert` methods _should_ have the verbose flag (and handle the verbose printing)


But further, I think that here we should be using the axis convert rather than the data convert... we don't want or need to convert constants or variables (the latter is not done anyway as we don't pass the kwarg) and we just transformed so there is only one axis.

ksunden:
But my point of "this is an implementation detail that does not need to be printed _ever_ stands

darienmorrow:
@ksunden I think I have done what you desired. If you desire something different, can you make the changes yourself?

ksunden (approved):

untzag (approved):

PR #999, opened 2021-04-21
From databroker

ksunden:
## Changes

Initial implementation of a from method to import data from databroker

currently does not put much info into attrs, nor does it handle units, since my example data did not have units stored to the data stream.

## Checklist

- [x] updated CHANGELOG.md
- [x] tests pass


WrightTools/data/_databroker.py
```diff
@@ -0,0 +1,37 @@
+from . import Data
```
ddkohler, line 1: `from ._data import Data` ?

untzag (approved):

untzag:
looks great @ksunden 

we know this is not ready for production, but getting it merged and released in this minimal form will allow us to work on other pieces of the pipeline

PR #1002, opened 2021-04-30
Explicitly store axes as fixed width numpy strings

ksunden:
Needed for compatibility with bluesky/tiled (via msgpack_numpy)

## Changes

## Checklist

- [x] updated CHANGELOG.md
- [x] tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Changed
 - complete units overhaul, now using pint library
+- explicitly store axes as fixed string dtype
 
 ### Fixed
 - Avoid passing both `vmin/vmax` and `norm` to `pcolor*` methods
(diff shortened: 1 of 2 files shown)


untzag:
@ksunden consider adding integration tests for tiled here?

kameyer226:
Guess this is an important decision here...maybe we should all sign off on this

ksunden:
@kameyer226 why do you think this is an important decision? this is one of the smallest effective changes I've proposed to any project recently. it really doesn't change all that much from the users point of view, just stores strings in a better way for cross language support and for integration with one particular library

ksunden:
No, it just means that all values shorter than the longest are null padded as stored, rather than being variable length as stored. There is almost no direct impact for users.

ksunden:
For reference, this is how we store all the other things like variable names and channel names

kameyer226 (approved):

PR #1035, opened 2021-11-02
artists.imshow

### LINKED ISSUE #1033: artists.imshow
It would be nice to have a WrightTools wrapper for matplotlib's `imshow`.  When rendering large, uniform arrays (e.g. camera images), `imshow` performance is noticably faster than pcolormesh. 

As a further embellishment, make `imshow` an option for plotting in interact2D (e.g. add a `uniform_spacing` kwarg).  

Additionally, one could apply non-linear ticks/tick labels to `imshow` to support non-uniform grids as well.

ddkohler:
## Changes

resolves #1033 .  

## Checklist

- [x] updated CHANGELOG.md
- [x] tests pass


WrightTools/artists/_base.py
```diff
@@ -20,7 +20,7 @@
 # --- define -------------------------------------------------------------------------------------
 
 
-__all__ = ["Axes", "Figure", "GridSpec", "apply_rcparams"]
+__all__ = ["Axes", "Figure", "GridSpec", "apply_rcparams", "_order_for_imshow"]
```
ksunden, line 23: I would tend to avoid adding a private method to `__all__`,

ksunden: you can still import it by name if you need it in another file, `__all__` just affects what gets brought into the namespace with `*` imports/just importing the module

kameyer226 (approved):

PR #1044, opened 2022-01-13
from_Solis - handle missing metadata on import

### LINKED ISSUE #986: from_solis:  import fails when no metadata
It is possible to export asc image from Solis software without metadata (clearly this is not preferred, but it should be handled gracefully).  

from_solis assumes "Data and Time" field exists in metadata (for attrs timestamp).  Fall back on file creation date as an alternative.

ddkohler:
## Changes

resolves #986 

## Checklist

- [x] updated CHANGELOG.md
- [x] tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+## Fixed
+- `data.from_Solis`: import works without metadata
+
+
 ## [3.4.3]
 
 ### Added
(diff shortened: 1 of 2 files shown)


darienmorrow (approved):

kameyer226:
Would time stamp and other metadata potentially be misread if passed between PC and Linux, say?  (Encoding)

ddkohler:
> Would time stamp and other metadata potentially be misread if passed between PC and Linux, say? (Encoding)

@kameyer226 based on my limited research, I think `st_mtime` is the most robust option.  I think date modified is consistent between machines.  Not perfect of course, but it gives a reasonable guess at the date the data was created.  

Another option for no metadata time is to pass `None` or equivalent as the timestamp...

kameyer226:
I think it is ok

PR #1050, opened 2022-03-04
`artists.scatter`: plot unstructured data

ddkohler:
## Changes

`wt.artists.scatter`: a convenience method for quickly viewing channels against two unstructured variables.

## Checklist

- [x] updated CHANGELOG.md
- [x] tests pass

### DIFF
--- a/WrightTools/artists/_base.py
+++ b/WrightTools/artists/_base.py
@@ -461,19 +461,20 @@ class Axes(matplotlib.axes.Axes):
             super().invert_yaxis()
         return out
 
-    def scatter_heatmap(self, *args, **kwargs):
+    def scatter(self, *args, **kwargs):
         """Scatter plot a channel against two _variables_.
         Scatter point color reflects channel values.
         Data need not be structured.
 
         Parameters
         ----------
-        data : 2D WrightTools.data.Data object
-            Data to plot.
-        x : Variable
-            azimuth variable
-        y : Variable
-            ordinate variable
+        data : 2D WrightTools.data.Data object (optional)
+            Data to plot.  If not provided, scatter reverts to 
+            [parent behavior](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.scatter.html?highlight=scatter#matplotlib.axes.Axes.scatter)
+        x : WrightTools.data.Variable
+            azimuth variable.
+        y : WrightTools.data.Variable
+            ordinate variable.
         channel : int or string (optional)
             Channel index or name. Default is 0.
         dynamic_range : boolean (optional)
@@ -495,16 +496,20 @@ class Axes(matplotlib.axes.Axes):
         matplotlib.collections.PathCollection
         """
         args = list(args)
-        data = args.pop(0)
-        x = data[args[0]][:].flatten()
-        y = data[args[1]][:].flatten()
-        args = [x, y] + args[2:]
-        channel = kwargs.pop("channel", 0)
-        channel_index = wt_kit.get_index(data.channel_names, channel)
-        z = data.channels[channel_index][:].flatten()
-        cmap = self._parse_cmap(data, channel_index=channel_index)["cmap"]
-        cmap = cmap(z)
-        return super().scatter(x, y, **kwargs, c=cmap)
+        if isinstance(args[0], Data):
+            data = args.pop(0)
+            x = data[args[0]][:].flatten()
+            y = data[args[1]][:].flatten()
+            args = [x, y] + args[2:]
+
+            channel = kwargs.pop("channel", 0)
+            channel_index = wt_kit.get_index(data.channel_names, channel)
+            cmap = self._parse_cmap(data, channel_index=channel_index, **kwargs)["cmap"]
+            z = data.channels[channel_index][:].flatten()
+            z = cmap(z)
+            kwargs["c"] = z
+        
+        return super().scatter(*args, **kwargs)
 
     def pcolormesh(self, *args, **kwargs):
         """Create a pseudocolor plot of a 2-D array.
--- a/WrightTools/artists/_base.py
+++ b/WrightTools/artists/_base.py
@@ -461,6 +461,53 @@ class Axes(matplotlib.axes.Axes):
             super().invert_yaxis()
         return out
 
+
+    def scatter_heatmap(self, *args, **kwargs):
+        """Scatter plot a channel against two _variables_.  
+        Scatter point color reflects channel values.  
+        Data need not be structured.
+        
+        Parameters
+        ----------
+        data : 2D WrightTools.data.Data object
+            Data to plot.
+        x : Variable
+            azimuth variable
+        y : Variable
+            ordinate variable
+        channel : int or string (optional)
+            Channel index or name. Default is 0.
+        dynamic_range : boolean (optional)
+            Force plotting of all contours, overloading for major extent. Only applies to signed
+            data. Default is False.
+        autolabel : {'none', 'both', 'x', 'y'}  (optional)
+            Parameterize application of labels directly from data object. Default is none.
+        xlabel : string (optional)
+            xlabel. Default is None.
+        ylabel : string (optional)
+            ylabel. Default is None.
+        **kwargs
+            matplotlib.axes.Axes.scatter__ optional keyword arguments.
+
+            __ https://matplotlib.org/api/_as_gen/matplotlib.pyplot.scatter.html
+
+        Returns
+        -------
+        matplotlib.collections.PathCollection
+        """
+        args = list(args)
+        data = args.pop(0)
+        x = data[args[0]][:].flatten()
+        y = data[args[1]][:].flatten()
+        args = [x,y] + args[2:]
+        channel = kwargs.pop("channel", 0)
+        channel_index = wt_kit.get_index(data.channel_names, channel)
+        z = data.channels[channel_index][:].flatten()
+        cmap = self._parse_cmap(data, channel_index=channel_index)["cmap"]
+        cmap = cmap(z)
+        return super().scatter(x, y, **kwargs, c=cmap)
+
+
     def pcolormesh(self, *args, **kwargs):
         """Create a pseudocolor plot of a 2-D array.
 
--- a/WrightTools/artists/_base.py
+++ b/WrightTools/artists/_base.py
@@ -468,10 +468,13 @@ class Axes(matplotlib.axes.Axes):
         Data need not be structured.
         If data object is not provided, scatter reverts to the [matplotlib parent method](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.scatter.html?highlight=scatter#matplotlib.axes.Axes.scatter)
 
-        Parameters
-        ----------
+        args
+        ---------
         data : 2D WrightTools.data.Data object
             Data to plot.
+
+        kwargs
+        ----------
         x : int or string (optional)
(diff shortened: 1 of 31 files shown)


ksunden:
my suggestion off initial impression:

just call it "scatter", which is more consistent with how the others (pcolor, pcolormesh, contourf and so on) in this class behave, but add handling of data arg/channel arg much like those other methods do

ksunden:
The idea of that is that if you do _not_ pass a data object or if you pass (or if you pass "c" explicitly) it behaves as the underlying scatter does, much like our versions of the other mpl functions

kameyer226:
nominally for help related purposes, is args listed as an ordered list of parameters, where it is assumed each one is the next element in args?  Would have thought to have used the word "args" in place of the word "parameters"

kameyer226 (approved):
seems ok

kameyer226 (commented):
Code scanning shows odd results, does this fall on submitter to correct?

ddkohler:
@kameyer226 I believe we call them parameters because the items can be specified in the function by their key or by their order of appearance.  In any case, the "parameter" designation persists throughout this package.  If you think arguments are not clearly, let's open a new issue and discuss it there?

Thanks for giving the review.  I did indeed find some of the docstring was not correct after looking closer.

ddkohler:
@kameyer226 tests have been passing--some older commits were failing which was something I fixed.

WrightTools/artists/_base.py
```diff
@@ -461,6 +462,73 @@ def imshow(self, *args, **kwargs):
...
+            limits = {}
+            limits = self._parse_limits(data=data, channel_index=channel_index, **limits)
+            norm = Normalize(**limits)
+
+            cmap = self._parse_cmap(data, channel_index=channel_index, **kwargs)["cmap"]
+
+            z = data.channels[channel_index][:].flatten()
+            z = norm(z)
+            z = cmap(z)
+            kwargs["c"] = z
```
ksunden, line 520: I would expect that if "c" is passed in, it would be respected rather than overwritten

WrightTools/artists/_base.py
```diff
@@ -461,6 +462,73 @@ def imshow(self, *args, **kwargs):
...
 
+    def scatter(self, *args, **kwargs):
+        """Scatter plot a channel against two _variables_.
+        Scatter point color reflects channel values.
+        Data need not be structured.
+        If data object is not provided, scatter reverts to the [matplotlib parent method](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.scatter.html?highlight=scatter#matplotlib.axes.Axes.scatter)
+
+        Parameters
+        ----------
+        data : 2D WrightTools.data.Data object
```
ksunden, line 473: should be documented as  positional only

WrightTools/artists/_base.py
```diff
@@ -461,6 +462,73 @@ def imshow(self, *args, **kwargs):
...
+            Data to plot.
+        x : int or string (optional)
```
ksunden, line 475: Remaining should be documented as key word only  (with some iffiness about x and y, as those _can_ be passed as positional, but according to current implementation  only in the "pass through" sense where we are not doing data things)

WrightTools/artists/_base.py
```diff
@@ -461,6 +462,73 @@ def imshow(self, *args, **kwargs):
...
+        if isinstance(args[0], Data):
+            data = args.pop(0)
+            x = kwargs.pop("x", 0)
+            x = wt_kit.get_index(data.variable_names, x)
+            x = data.variables[x].full.flatten()
+
+            y = kwargs.pop("y", 1)
+            y = wt_kit.get_index(data.variable_names, y)
+            y = data.variables[y].full.flatten()
+            args = [x, y] + args[2:]
```
ksunden, line 506: why args[2:]?

one element is removed by pop on line 498, so you are skipping two args

WrightTools/artists/_base.py
```diff
@@ -461,6 +462,73 @@ def imshow(self, *args, **kwargs):
...
+        Returns
+        -------
+        matplotlib.collections.PathCollection
+        """
+        args = list(args)
+        if isinstance(args[0], Data):
+            data = args.pop(0)
+            x = kwargs.pop("x", 0)
+            x = wt_kit.get_index(data.variable_names, x)
+            x = data.variables[x].full.flatten()
```
ksunden, line 501: full may not be _quite_ what we want here (we want it expanded to match the full shape of the variables and channel used and then flattened)

WrightTools/artists/_base.py
```diff
@@ -461,6 +462,73 @@ def imshow(self, *args, **kwargs):
...
+
+        Returns
+        -------
+        matplotlib.collections.PathCollection
+        """
+        args = list(args)
+        if isinstance(args[0], Data):
+            data = args.pop(0)
+            x = kwargs.pop("x", 0)
+            x = wt_kit.get_index(data.variable_names, x)
```
ksunden, line 500: I would expect axes, not variables for the default/index lookup

ksunden: possibly with fallback to variable (or even channel....)

tests/artists/test_scatter.py
```diff
@@ -0,0 +1,35 @@
...
+    data.create_variable("y", values=4 * np.pi * (rng2 - 0.5))
+    data.create_channel("z", values=np.sin((data.x[:] ** 2 + data.y[:] ** 2) ** 0.5), signed=True)
+    fig, gs = wt.artists.create_figure()
+    ax = plt.subplot(gs[0])
+    ax.scatter(data, "x", "y", channel="z", s=20, alpha=0.5)
+    data.close()
+
+
+if __name__ == "__main__":
+    t0 = test_scatter_broadcast_vars()
```
ksunden, line 33: remove saving to variable, it is unused and always None

tests/artists/test_scatter.py
```diff
@@ -0,0 +1,35 @@
...
+def test_scatter_broadcast_vars():
+    data = wt.data.Data(name="data")
+    rng1 = np.random.rand(50)[:, None]
+    rng2 = np.random.rand(50)[None, :]
+    data.create_variable("x", values=2 * rng1 - 1)
+    data.create_variable("y", values=2 * rng2 - 1)
+    data.create_channel("z", values=np.exp(-(data.x[:] ** 2 + data.y[:] ** 2)))
+    fig, gs = wt.artists.create_figure()
+    ax = plt.subplot(gs[0])
+    ax.scatter(data, "x", "y", channel="z", s=20, alpha=0.5)
```
ksunden, line 15: These calls don't do what you think... if you swapped x and y in the function call, current implementation would not flip your axes, I believe

tests/artists/test_scatter.py
```diff
@@ -0,0 +1,35 @@
+import numpy as np
+import WrightTools as wt
+import matplotlib.pyplot as plt
+
+
+def test_scatter_broadcast_vars():
+    data = wt.data.Data(name="data")
+    rng1 = np.random.rand(50)[:, None]
```
ksunden, line 8: if you are going to use random, please set the seed so tests are reproducible

WrightTools/artists/_base.py
ddkohler: ok, I think I get it.  `data` is pulled as the first element of `*args`, so it has to be positional.   @kameyer226 you had the right idea.

Is this a correction that extends to all our plot functions?

ksunden (changes requested):
See comments linked to lines of code

WrightTools/artists/_base.py
ksunden: if you want x and y to be passable as positional args, they need to handled as such (it appears you do based on your tests)

WrightTools/artists/_base.py
ksunden: Yeah, I think it does need to apply to other plot functions... most of them use `_parse_plot_args` which also expects data to be positional and more or less everything else to be kwargs (with additional args being passed as positional arguments to the underlying function after the x/y[/z] which are extracted from the data object

WrightTools/artists/_base.py
ksunden: Essentially the premise is "if I explicitly ask you to do something, I expect you to do it" this includes "I _just_ want a pure scatter plot of x vs y with black points, so I will pass `c="k"`"

ksunden:
> nominally for help related purposes, is args listed as an ordered list of parameters, where it is assumed each one is the next element in args? Would have thought to have used the word "args" in place of the word "parameters"

Matplotlib (as we are interfacing/emulating here) is particularly bad about allowing things like "the number of arguments is variable and their types matter and we just want to do the thing you want to do as flexibly as possible"

The down side of this is that the lines get fuzzy here

For "normal" python functions, most arguments (or parameters, these are equivalent terms for this discussion) can be used in either a positional sense or in a keyword sense.
There exist ways of having named arguments that appear in your function signature that are "keyword only" or "positional only".

Here we kind of give up on enumerating all of the arguments that are possible and just say `*args, **kwargs` which is "we will deal with args as a sequence and kwargs as a dictionary"
This means that we either need to check both places if we want things to behave as "normal" functions do or we should be specific about what is expected to be "positional" vs "keyword"

WrightTools/artists/_base.py
kameyer226: I like the use of "parameters', 'args', and 'kwargs', but I do not know if my likes are compatible with the community here

WrightTools/artists/_base.py
```diff
@@ -503,7 +506,7 @@ def scatter(self, *args, **kwargs):
             y = kwargs.pop("y", 1)
             y = wt_kit.get_index(data.variable_names, y)
             y = data.variables[y].full.flatten()
-            args = [x, y] + args[2:]
+            args = [x, y] + args[1:]
```
ksunden, line 509: remember that you pop data out of args

WrightTools/artists/_base.py
ddkohler: I have realized this is more complex to implement if we allow users to specify parent kwarg "c".  This is because the "c" in it's raw form is sequential (1D).  Since we default to a channel, I want to say it is implied here that a channel will determine the coloring of scatter points.  So I will remove the option to specify parent kwarg "c".  Users can instead use "cmap".

kameyer226:
I didn't approve yet because reviewer @ksunden  made a point about the pops removing elements that I didn't see as resolved.  See Mar 9

WrightTools/artists/_base.py
```diff
@@ -463,6 +464,100 @@ def imshow(self, *args, **kwargs):
             super().invert_yaxis()
         return out
 
+    def scatter(self, *args, **kwargs):
+        """Scatter plot a channel against two _variables_.
+        Scatter point color reflects channel values.
+        Data need not be structured.
+        If data object is not provided, scatter reverts to the [matplotlib parent method](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.scatter.html?highlight=scatter#matplotlib.axes.Axes.scatter)
```
ksunden, line 471:
```suggestion
        If data object is not provided, scatter reverts to the `matplotlib parent method <https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.scatter.html?highlight=scatter#matplotlib.axes.Axes.scatter>`_.
```

Docstrings use [rst links](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#hyperlinks), not markdown links

ksunden (approved):

PR #1051, opened 2022-03-09
from_LabRAM

ddkohler:
## Changes

`from_LabRAM`: import data from Horiba LabRAM HR imaging spectrometer.  Thanks to Chris Roy for writing the bulk of this method.

@ksunden related note:  this imaging spectrometer outputs `.l6m` binary files (I believe this stands for [labspec 6](https://www.horiba.com/int/products/detail/action/show/Product/labspec-6-spectroscopy-suite-software-1843/). I am not sure how similar they are to the Aramis `.ngc` files.  Do you have opinions on how to work with this?  No urgency, as txts are available.

## Checklist

- [x] add tests for 1D, 2D, 3D datasets
- [x] finish sorting attrs
- [x] updated documentation
- [x] updated CHANGELOG.md
- [x] tests pass


ksunden:
If you can get me a few examples (including some with text output as well to compare/know what I'm looking for) I'm happy to take a look at the binary files... actually been feeling the itch to do that again... We can try to put them through the NGC parser, can't hurt, but chances are the extension is different for a reason...

ksunden:
That said, ngc is what labspec5 used, so maybe they just wanted to change the extension...

ksunden:
consider reading units from the bottom of the metadata where there is an explicit "Axis_units[n]" entry in the metadata rather than assuming it is the same as the earlier metadata/constants

ddkohler:
> consider reading units from the bottom of the metadata where there is an explicit "Axis_units[n]" entry in the metadata rather than assuming it is the same as the earlier metadata/constants

I tried, but that metadata is not included in some 1D and 2D scans, so it's not a complete solution.  There is definitely a smart way to use those axis entries, but it would require more time to develop the strategy

kameyer226 (approved):

PR #1052, opened 2022-03-09
Fix errors identified by CodeQL

ksunden:
Mostly things like more explicit error messages when unexpected values/types are passed instead of waiting for it to fail on something like NameError for a variable that was not in the users code

## Checklist

- [x] updated CHANGELOG.md
- [x] tests pass


kameyer226:
CodeQL is not working right, the code was not in error.   Can this get reported to CodeQL curators?

ksunden:
@kameyer226 I think you'll find that it is indeed exactly correct about to fix it

kameyer226 (approved):

Issue #1065, opened 2022-05-16
Test wt.data methods with complex channels


kameyer226:
Suggest use of complex arrays for test channels in the WrightTools tests for all methods that should be able to modify complex arrays.   This can facilitate use of WrightSim objects for methods in WrightTools once converted to a WrightTools object.   Some kind of handling will be necessary for use of artists.quickNd  with complex arrays.

kameyer226:
_pycmds.py line 168  
`values = np.full(np.prod(shape), np.nan)`
may want some dtype argument for complex values but I do not know how to incorporate it.

ddkohler:
I will note it makes more sense to me to e.g. store real and imaginary parts as separate channels.  We don't have support plotting of complex numbers.  Complex channels will not be great as splitting expressions, either.  That said, I don't see a reason why storing a complex channel shouldn't work, so long as you don't try to do too much with it...

ddkohler:
> _pycmds.py line 168 `values = np.full(np.prod(shape), np.nan)` may want some dtype argument for complex values but I do not know how to incorporate it.

I think we can assume pycmds data does not have complex values, so I'm not worried here.

kameyer226:
I haven't validated the calculation of moments with complex numbers to be sure that the formulas there work,  splines of complex numbers must be fun...   I don't think it necessary to force methods to plot complex values, I think we can leave that incumbent on the user.

kameyer226:
Also, validated the wt5 saves and loads work

ddkohler:
@kameyer226 was this resolved in #1067  ?

PR #1067, opened 2022-05-19
Fix complex array operations

kameyer226:
## Changes

fixes to `data.split` and `data.join`,  added tests, and verified .wt5 saves and opens work with complex arrays
(np.complex128 tested only)

## Checklist

- [X ] added tests, if applicable
- [ ] updated documentation, if applicable  N/A
- [ X] updated CHANGELOG.md
- [ X] tests pass

### DIFF
--- a/WrightTools/data/_data.py
+++ b/WrightTools/data/_data.py
@@ -861,8 +861,10 @@ class Data(Group):
                 require_kwargs["dtype"] = np.dtype(np.float64)
             else:
                 require_kwargs["dtype"] = dtype
-            if require_kwargs["dtype"].kind in "fcmM":
+            if require_kwargs["dtype"].kind in "fmM":
                 require_kwargs["fillvalue"] = np.nan
+            elif require_kwargs["dtype"].kind in "c":
+                require_kwargs["fillvalue"] = np.complex128(np.nan)
             else:
                 require_kwargs["fillvalue"] = 0
         else:
@@ -917,8 +919,10 @@ class Data(Group):
                 shape = self.shape
             if dtype is None:
                 dtype = np.dtype(np.float64)
-            if dtype.kind in "fcmM":
+            if dtype.kind in "fmM":
                 fillvalue = np.nan
+            elif dtype.kind in "c":
+                fillvalue = np.complex128(np.nan)
             else:
                 fillvalue = 0
         else:
@@ -1716,7 +1720,7 @@ class Data(Group):
                 out_arr = np.full(omask.shape, np.nan)
                 imask = wt_kit.enforce_mask_shape(imask, var.shape)
                 out_arr[omask] = var[:][imask]
-                out[i].create_variable(values=out_arr, **var.attrs)
+                out[i].create_variable(values=out_arr, **var.attrs, dtype=var.dtype)
 
         for ch in self.channels:
             for i, (imask, omask, cut) in enumerate(zip(masks, omasks, cuts)):
@@ -1725,10 +1729,10 @@ class Data(Group):
                     continue
                 omask = wt_kit.enforce_mask_shape(omask, ch.shape)
                 omask.shape = tuple([s for s, c in zip(omask.shape, cut) if not c])
-                out_arr = np.full(omask.shape, np.nan)
+                out_arr = np.full(omask.shape, np.nan, dtype=ch.dtype)
                 imask = wt_kit.enforce_mask_shape(imask, ch.shape)
                 out_arr[omask] = ch[:][imask]
-                out[i].create_channel(values=out_arr, **ch.attrs)
+                out[i].create_channel(values=out_arr, **ch.attrs, dtype=ch.dtype)
 
         if verbose:
             for d in out.values():
--- a/WrightTools/kit/_array.py
+++ b/WrightTools/kit/_array.py
@@ -237,7 +237,9 @@ def share_nans(*arrs) -> tuple:
     list
         List of nD arrays in same order as given, with nan indicies syncronized.
     """
-    nans = np.zeros(joint_shape(*arrs))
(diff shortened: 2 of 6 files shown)


kameyer226:
Closes #1065

WrightTools/data/_data.py
```diff
@@ -861,8 +861,10 @@ def create_channel(
                 require_kwargs["dtype"] = np.dtype(np.float64)
             else:
                 require_kwargs["dtype"] = dtype
-            if require_kwargs["dtype"].kind in "fcmM":
+            if require_kwargs["dtype"].kind in "fmM":
                 require_kwargs["fillvalue"] = np.nan
+            elif require_kwargs["dtype"].kind in "c":
+                require_kwargs["fillvalue"] = np.complex128(np.nan)
```
ddkohler, line 867: I think you want to fill with `np.complex(np.nan, np.nan)`.  

Comment applies to filling all blank arrays.

tests/data/split.py
```diff
@@ -33,6 +33,25 @@ def test_split():
...
+def test_split_complexarray():
+    p = datasets.PyCMDS.wm_w2_w1_000
+    a = wt.data.from_PyCMDS(p)
+    a.create_channel(name="complex", values=np.complex128(a.channels[0][:]), dtype=np.complex128)
+    exprs = a.axis_expressions
+    split = a.split(0, [19700], units="wn")
+    assert exprs == a.axis_expressions
+    assert exprs == split[0].axis_expressions
+    assert len(split) == 2
+    print(split[0].shape)
```
ddkohler, line 45: remove print

tests/data/split.py
```diff
@@ -33,6 +33,25 @@ def test_split():
     split.close()
 
 
+def test_split_complexarray():
+    p = datasets.PyCMDS.wm_w2_w1_000
+    a = wt.data.from_PyCMDS(p)
+    a.create_channel(name="complex", values=np.complex128(a.channels[0][:]), dtype=np.complex128)
+    exprs = a.axis_expressions
+    split = a.split(0, [19700], units="wn")
+    assert exprs == a.axis_expressions
```
ddkohler, line 42: unnecessary assert statement

WrightTools/kit/_array.py
```diff
@@ -237,7 +237,15 @@ def share_nans(*arrs) -> tuple:
     list
         List of nD arrays in same order as given, with nan indicies syncronized.
     """
-    nans = np.zeros(joint_shape(*arrs))
+    typ = int
```
ddkohler, line 240: Consider simpler syntax e.g.
```
typ = np.complex128 if "c" in [arr.dtype.kind for arr in arrs] else arrs[0].dtype
```

Note that `np.nan * complex(0) = nan + nanj`, so no concern here about 0-valued imaginary parts

WrightTools/kit/_array.py
```diff
@@ -237,7 +237,15 @@ def share_nans(*arrs) -> tuple:
...
+    for arr in arrs:
+        if arr.dtype.kind in "fmM":
+            if typ != np.complex128:
+                typ = arr.dtype
+        if arr.dtype.kind in "c":
+            typ = arr.dtype
+
+    nans = np.zeros(joint_shape(*arrs), dtype=typ)
     for arr in arrs:
         nans *= arr
```
ddkohler, line 250: Note that `np.nan * complex(0) = nan + nanj`, so no concern here about 0-valued imaginary parts .  This step should work

WrightTools/kit/_array.py
kameyer226: what happens if the 2nd array is float and 1st is int?   I was considering having to broadcast into the most "np.nan"-like of all datatypes, so the hierarchy was int < float < complex   and that any float would require an int to broadcast to float.

ddkohler: That's a fair point.  I will note that none of these array creation code blocks you have revised really think about integer types (`dtype.kind`=`u`,`i`).  Consideration of integer types might be worth looking into as a separate issue?  e.g. quite a few more tests should be implemented to really tackle the int type...

ddkohler: The good news is that casting ints as floats doesn't have dramatic consequences for our typical usage.   I can't think of an example where this would really be a problem...

kameyer226: somehow the original code was able to handle it, yet I couldn't understand how...a simple line of code would be nicer though

WrightTools/data/_data.py
```diff
@@ -917,8 +919,10 @@ def create_variable(
                 shape = self.shape
             if dtype is None:
                 dtype = np.dtype(np.float64)
-            if dtype.kind in "fcmM":
+            if dtype.kind in "fmM":
```
ddkohler, line 922: Actually, now that I think about this, does dtype "m", "M" support `np.nan`?
```
x = np.zeros((3), dtype=np.timedelta64)
x[0] = np.nan
```
This throws an error...fill values for time should be NaT?

WrightTools/kit/_array.py
ddkohler: To keep this hierarchy and make the code more readable, consider an elif structure:
```
kinds = {arr.dtype.kinds for arr in arrs}
if "c" in kinds:
    dtype = complex
elif "f" in kinds:
   dtype = float
elif "i" in kinds:
   dtype = int
else:
   dtype = arrs[0].dtype
```

WrightTools/data/_data.py
kameyer226: Since I do not know what kind of data arrays we use require "m" kind for our purposes, I took that kind out

kameyer226:
Hi, all conversations resolved.

Issue #1062, opened 2022-04-16
artists colormaps: new normalizations


ddkohler:
Matplotlib now has good support for colormap normalization strategies that we have implemented internally (e.g. through dynamic range).  

https://matplotlib.org/stable/tutorials/colors/colormapnorms.html

One minor benefit is that this would make behind-the-scenes code simpler (e.g. artists.base._parse_plot_args), but it should also make it easy for other normalizations that we have been wanting (e.g. TwoSlopeNorm).

ddkohler:
This issue is related to #38

PR #1073, opened 2022-06-09
interact2d cmap options

ddkohler:
## Changes

* __Bump versions: mpl >= 3.4, Python >=3.7__

interact2D
* Accept cmap as kwarg.
* Signed channels have signed cmap as default.
* Implement matplotlib Norm objects.
* Addresses #1062 

## Checklist

- [x] updated CHANGELOG.md
- [x] tests pass


ksunden:
You will actually need to get rid of py3.6 if you want to bump mpl version...


Like I said, I'm good with it, but yeah, mpl itself has removed support for 3.6

ddkohler:
Would be interesting to have [TwoSlopeNorm](https://matplotlib.org/stable/api/_as_gen/matplotlib.colors.TwoSlopeNorm.html), but I think that is the less desired than `CenteredNorm` for signed data.

We could introduce yet another kwarg to interact2D to make this possible...

kameyer226 (approved):
Cool!

PR #1074, opened 2022-06-09
`Data.translate_to_txt`

### LINKED ISSUE #1071: Support export to csv
There is still demand within the user group to have an easy way to export wt5 data to work with e.g. Excel.  

This could be a Data method (`data.to_csv(filename)`), could also be an entry-point?  e.g. `wt-to-csv --filename data.wt5`.

ddkohler:
## Changes

* added Data method `translate_to_txt` for serialized, text-based data representation for operability with text-based tools.
* resolves #1071 .

## Checklist

- [x] verbose kwarg
- [x] updated CHANGELOG.md
- [x] tests pass


ddkohler:
Pausing development for now.  Currently looking for input from users who want this feature, to help guide the scope of this function.

ddkohler:
In the process of testing this with John.  We will hold off merges until it passes the "John" test.  In the meantime, open to review from any interested parties.

darienmorrow:
@ddkohler Could we put the `sys.stdout.` calls behind an `if verbose:` kwarg gate?

ddkohler:
@darienmorrow good idea

darienmorrow:
As a reminder, IRL we tossed around the idea of using [`np.savetxt`](https://numpy.org/doc/stable/reference/generated/numpy.savetxt.html) to speed up writing instead of writing line by line.

darienmorrow:
What is the status of this PR @ddkohler? Did John get back to you?

ddkohler:
@darienmorrow I think the general opinion is that the format is good enough.  If it's important to you, can you test and see if it's adequate for of your applications?  Just added chunkwise writing (similar to the savetxt idea), so there should be a speedup.

Failing tests are non-stochastic and unrelated--my believe they are consequences of #1112 .  We can fix those in another PR.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 ## [3.4.5]
 
 ### Added
+- `Data.translate_to_txt`: serialize channels and variables and write as a text file.
```
ksunden, line 25: This should be moved up in this file

WrightTools/data/_data.py
```diff
@@ -1928,6 +1929,94 @@ def transform(self, *axes, verbose=True):
...
+        * wt5, the native file format for Data objects, is a specific variant
+        of the HDF5 file format. HDF5 has well-developed tools for working with
+        generic datasets.  See
+        https://portal.hdfgroup.org/display/knowledge/How+to+convert+an+HDF4+or+HDF5+file+to+ASCII+%28text%29+or+Excel
+        for more information.
+        """
+        # attrs
+        import tidy_headers
+
+        tidy_headers.write(filepath, {k: v for k, v in self.attrs.items()})
```
ksunden, line 1976:
```suggestion
        tidy_headers.write(filepath, self.attrs)
```

I think this should work

tests/data/translate_to_txt.py
```diff
@@ -0,0 +1,27 @@
...
+import pathlib
+import pytest
+
+import WrightTools as wt
+from WrightTools import datasets
+
+# --- tests ---------------------------------------------------------------------------------------
+
+
+@pytest.mark.skip("don't want to write files")
```
ksunden, line 16: I would use https://docs.python.org/3/library/tempfile.html?highlight=tempfile#tempfile.NamedTemporaryFile to both clean up after the test and actually test it rather than just skipping

ksunden (changes requested):

tests/data/translate_to_txt.py
```diff
@@ -0,0 +1,28 @@
...
+from tempfile import NamedTemporaryFile
+
+
+# --- tests ---------------------------------------------------------------------------------------
+
+
+def test_datasets_mos2():
+    d = wt.open(datasets.wt5.v1p0p1_MoS2_TrEE_movie).at(w2=[18000, "wn"])
+    tmp = NamedTemporaryFile(delete=False)
+    d.translate_to_txt(tmp.name, verbose=True)
```
ksunden, line 20:
```suggestion
    with NamedTemporaryFile(delete=False) as name:
        d.translate_to_txt(tmp.name, verbose=True)
```

Slightly more correct as it will ensure that cleanup happens (because context managers _always_ run `__exit__`.

darienmorrow (approved):

PR #1075, opened 2022-06-10
fft: satisfy Plancherel

### LINKED ISSUE #1076: kit.fft: fourier coefficients ignore x-axis spacing
kit fft expands on np.fft.fft by accepting an x-axis as an argument to give convenient units for Fourier space.  It neglects to use these units to calculate the Fourier transform itself (e.g. we calculate `array.sum()` instead of `array.sum() * dt`).  Since our method we explicitly call for x-units, we should use their to calculate the Fourier coefficients as well.

This comes up when you want to relate spectral amplitudes to temporal features.  e.g.
```
    t = np.linspace(-10, 10, 10000)
    z = np.sin(2 * np.pi * t)
    wi, zi = wt.kit.fft(t, z)
    intensity_time = (z**2).sum() * (t[1] - t[0])
    intensity_freq = np.abs(zi).sum() * (wi[1] - wi[0])
```
By Parseval's (or Plancherel's) theorem, both `intensity_time` and `intensity_freq` should be equal to within calculation error.  This is not currently the case.  In general, all our FT coefficients are off by a constant that depends on the spacing between points.

ddkohler:
## Changes

Spectral intensity across all colors should be equal to intensity across all time.
Resolves #1076 .

## Checklist

- [x] updated CHANGELOG.md
- [x] tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Fixed
+- `kit.fft`: fixed bug where Fourier coefficients were off by a scalar factor.
+
 ## [3.4.4]
 
 ### Added
--- a/WrightTools/kit/_array.py
+++ b/WrightTools/kit/_array.py
@@ -131,7 +131,7 @@ def fft(xi, yi, axis=0) -> Tuple[np.ndarray, np.ndarray]:
     Parameters
     ----------
     ti : 1D numpy.ndarray
-        Independent variable specifying data coordinates. Must be monotonic, 
+        Independent variable specifying data coordinates. Must be monotonic,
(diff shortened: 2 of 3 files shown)


darienmorrow (approved):

ksunden:
https://numpy.org/doc/stable/reference/generated/numpy.fft.fft.html#numpy.fft.fft

Somewhat curious if simply passing `norm="ortho"` has the same effect?

ddkohler:
> https://numpy.org/doc/stable/reference/generated/numpy.fft.fft.html#numpy.fft.fft
> 
> Somewhat curious if simply passing `norm="ortho"` has the same effect?

No, I think their norm argument relates to unitary vs. non-unitary definitions. https://numpy.org/doc/stable/reference/routines.fft.html#normalization

Note that np.fft.fft cannot incorporate this feature because it does not accept x-axis/spacing as an argument.

Issue #697, opened 2018-08-03
advanced data views?


p770193:
Some crucial functions of the data object focus on selecting/redistributing our data (e.g., `split` and `chop`).  

These methods have strong parallels with slicing and advanced indexing of the `numpy.ndarray` class.
One feature of ndarray slicing/indexing is that you get views/references of the structured data, not new arrays. Can/should we do the analogous thing when using some of these methods?  For example, we could introduce the function `view` :
```
>>> data.axes, data.shape
(<..."d1">, <..."d2"...>, <..."w1"...>, <..."w2"...>), (51, 51, 21, 21)
>>> data.view('w1', 'w2', 'd2', at={["d1": [500, 'fs']]})
>>> data.axes, data.shape
(<..."d2"...>, <..."w1"...>, <..."w2"...>), (51, 51, 21)
>>> data.view('all')
>>> data.axes, data.shape
(<..."d1">, <..."d2"...>, <..."w1"...>, <..."w2"...>), (51, 51, 21, 21)
```
Or something like this.  The point is that the data object knows which data we want to see, but doesn't recopy all that data to a new file/object.  If we actually wanted to manipulate these subsets of the data, we would copy the data object first (as we would do with subsets of numpy arrays).  This would remove the overhead of passing on duplicate axes, variables, channels, and attributes that are highly redundant.

If a data object works like a numpy array, I tend to feel like there are usability/simplicity benefits.  Please give thoughts.

ksunden:
This is an interesting Idea, I want to link this idea to the idea of #9 (Our oldest extant issue)

I'm going to say that this is a relatively low priority task for right now, but good to have some thoughts floating about it.

PR #1078, opened 2022-06-14
easier chopping/slicing operations on Data

### LINKED ISSUE #9: slicing into data objects
Would be really cool to get this done. We're ready to tackle the task, someone just needs to sit down and figure out how. 

Idea is to be able to just enter

``` python
data[5:, -1]
```

using [numpy style indexing](http://docs.scipy.org/doc/numpy/reference/arrays.indexing.html). Both data.zis and data.axes would be updated.

Potentially relevant: http://stackoverflow.com/questions/22632937/how-to-make-numpy-overloading-of-add-independent-on-operand-order

Also: http://dimarray.readthedocs.org/en/latest/_notebooks_rst/tutorial.html#data-structure

### LINKED ISSUE #956: using split in units of index values
Idea is to use kwarg units="ind", or some other novel approach, to enable split to be performed at fixed index values of the data array rather than in prescribed units.   Reason is to eliminate ragged edge behavior that occasionally manifests in > 1D arrays, which creates  (probably necessary) 'nan' entries in the axes fields of the resultant splits and disables further data arithmetic such as moments.

Specific example:  say w2,wa data is transformed to w2,wa-w2.  Then one may want to split data to create a narrow window around 0 in wa-w2.   This could lead to ragged edges of the split due to the transform.

### LINKED ISSUE #1068: new Data method `at` for simpler chop syntax
```
x = data.at(w2=[500, "nm"])
y = data.chop("w1", at={"w2":[500, "nm])[0]
```
Create `at` such that `x==y`

Also: refactor chop so `at` is the worker function for `chop`.

ddkohler:
## Changes

* new Data method `at` for simpler syntax of chopping with `at` kwarg.  This implementation involved refactoring `Data.chop` into more modular code so it could be reused.
* `Data.__getitem__` now supports simple slicing of underlying arrays (`sliced_data = data[:5, 1]`).
* resolves #9 (does not allow getitem with floats, however)
* addresses, but does not resolve, #697 (methods implemented here do not offer "views," only copies of the original data arrays)
* resolves #956 
* resolves #1068 

## Checklist

- [x] updated CHANGELOG.md
- [x] tests pass


WrightTools/data/_data.py
```diff
@@ -304,23 +304,57 @@ def bring_to_front(self, channel):
...
+        See Also
+        --------
+        Data.chop : also reduces dimensionality, but returns a collection of one or more
+        data objects. Kept axes can be sliced into a collection
+        """
+        if idx is None:
+            idx = np.array([slice(None)] * len(self._axes))
         for axis, point in at.items():
+            # DDK: is this stuff necessary?  axis must be str says docs...
+            #   this is some hidden way to deal with Axis indices?
```
ksunden, line 341: so, axis is a string because we use `**kwargs` (we call it `at`, but same difference)


I think there may technically be implementations of python that allow non-string kwarg keys to be passed in in certain circumstances (e.g. doing `foo(**{1:3})`) but CPython disallows it so probably not a thing we need to actually account for.

I feel like I remember passing as ints in the past, but that may have been an older implementation that passed a dict rather than as kwargs. (even py 2.7 explicitly disallows kwargs that aren't strings)

ddkohler:
Still have to add tests, but I wanted to open this up for review so people can try the new features/test this branch on existing projects. These changes would benefit from a diverse range of testing, so it's preferable to have several users play around with this before merging.

ksunden (changes requested):
Add tests :stuck_out_tongue_closed_eyes: 

While that is a true request, partly putting this in to prevent big green button once only one person approves since this is the style of change that we want multiple reviews on.

WrightTools/data/_data.py
```diff
@@ -365,100 +414,125 @@ def chop(self, *args, at=None, parent=None, verbose=True) -> wt_collection.Colle
...
+            idx[at_axes] = at_idx[at_axes]
+            data = self._from_slice(idx, name=name, parent=out)
         out.flush()
         # return
         if verbose:
-            print("chopped data into %d piece(s)" % len(out), "in", new_axes)
+            print("chopped data into %d piece(s)" % len(out), "in", out[0].axis_expressions)
+        return out
+
+    def __getitem__(self, key) -> object:
```
ddkohler, line 469: [Type hints for Python 3.6 are significantly different from Python 3.7+](https://stackoverflow.com/questions/33533148/how-do-i-type-hint-a-method-with-the-type-of-the-enclosing-class); for this reason I cannot easily tell hint this as a Data object.  

Does anyone have objections to dumping support for Python 3.6?  Note that other scientific packages do not feel it is important to keep support for 3.6 https://numpy.org/neps/nep-0029-deprecation_policy.html

ksunden: Though I will note that I think you should be able to use forward reference type hint (i.e. just put it in quotes)

ksunden:
Current behavior does not support indexing into multiple dimensions:

```
>>> data[0::2, :3, :2]
│ <stdin>:1 in <module>                                                                            │
│ /home/kyle/src/wright-group/WrightTools/WrightTools/data/_data.py:473 in __getitem__             │
│                                                                                                  │
│    470 │   │   if type(key) in [int, slice]:                                                     │
│    471 │   │   │   return self._from_slice(key)                                                  │
│    472 │   │   elif (type(key) == tuple) and np.all([type(ki) in [int, slice] for ki in key]):   │
│ ❱  473 │   │   │   return self._from_slice(*key)                                                 │
│    474 │   │   else:                                                                             │
│    475 │   │   │   return super().__getitem__(key)                                               │
│    476                                                                                           │
│                                                                                                  │
│ /home/kyle/src/wright-group/WrightTools/WrightTools/data/_data.py:505 in _from_slice             │
│                                                                                                  │
│    502 │   │   if parent is None:                                                                │
│    503 │   │   │   out = Data(name=name)                                                         │
│    504 │   │   else:                                                                             │
│ ❱  505 │   │   │   out = parent.create_data(name=name)                                           │
│    506 │   │                                                                                     │
│    507 │   │   for v in self.variables:                                                          │
│    508 │   │   │   kwargs = {}                                                                   │

```
as the key, a tuple, is passed with the expansion operator, but the private `_from_slice` method does rather expects the tuple itself to be passed, removing the `*` on line 473 appears to correct this

ksunden:
Outside of that one character change above, I believe this behaves as expected and is ready to be put out into the world.

WrightTools/data/_data.py
```diff
@@ -365,100 +411,128 @@ def chop(self, *args, at=None, parent=None, verbose=True) -> wt_collection.Colle
...
+        return out
+
+    def __getitem__(self, key) -> Data:
+        """
+        data[5, :3]; return new data object with those array slices
+        """
+        if type(key) in [int, slice]:
+            return self._from_slice(key)
+        elif (type(key) == tuple) and np.all([type(ki) in [int, slice] for ki in key]):
+            return self._from_slice(*key)
```
ksunden, line 473:
```suggestion
            return self._from_slice(key)
```

ksunden (changes requested):

ksunden (approved):

ksunden:
I may have liked to see something added to docs, but honestly this is a power user feature at least for now, so the autogenerated api docs are probably good enough for now

PR #1080, opened 2022-06-22
ensure artists.quickND plotters always close chopped data files

ksunden:
## Changes

Wrap the implementations of quick1D and quick2D with a closing context manager for the chopped collection

## Checklist

- [x] added tests, if applicable... I wish testing this in pytest was easier, but it would rely on some weirdness that I don't know if I want to rely on
- [x] updated CHANGELOG.md
- [x] tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Fixed
 - `kit.fft`: fixed bug where Fourier coefficients were off by a scalar factor.
+- ensure `artists.quickND` plotters always close chopped data files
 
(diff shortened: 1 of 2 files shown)


ksunden:
Unfortunately this did not actually fix the behavior I was trying to address, though it remains a good change that should be included in my opinion (just not the silver bullet I hoped it'd be)

ddkohler (approved):
I don't have a great way to test this code other than verifying quicknD still works for me (which it seems to).  lgtm

ksunden:
The diff is large, but only because of whitespace, its basically just ensuring that data objects which get opened by quick functions are closed explicitly and don't just hang out waiting

PR #1085, opened 2022-07-19
enable compression

### LINKED ISSUE #422: dataset compression
shall we compress our channels and variables? (I think yes)

if so, how?

[h5py offers 3 lossless compression filters](http://docs.h5py.org/en/latest/high/dataset.html#lossless-compression-filters)

I think we should use GZIP, the most universal compression option we have available

ksunden:
## Changes

Enable passing compression/compression opts/shuffle arguments to h5py `require_dataset` 

Initially this is posted to allow testing in bluesky-in-a-box, so draft without any tests/docs added yet

Closes #422 

## Checklist

- [x] updated CHANGELOG.md
- [x] tests pass

### DIFF
--- a/WrightTools/data/_variable.py
+++ b/WrightTools/data/_variable.py
@@ -34,7 +34,6 @@ class Variable(Dataset):
             Additional keys and values to be written into dataset attrs.
         """
         self._parent = parent
-        super_kwargs = {k: kwargs[k] for k in ("compression", "shuffle") if k in kwargs}
         super().__init__(id)
         if units is not None:
             self.units = units
--- a/WrightTools/data/_variable.py
+++ b/WrightTools/data/_variable.py
@@ -34,6 +34,7 @@ class Variable(Dataset):
             Additional keys and values to be written into dataset attrs.
         """
         self._parent = parent
+        super_kwargs = {k: kwargs[k] for k in ("compression", "shuffle") if k in kwargs}
         super().__init__(id)
         if units is not None:
             self.units = units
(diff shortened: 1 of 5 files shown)


WrightTools/data/_variable.py
```diff
@@ -34,6 +34,7 @@ def __init__(self, parent, id, units=None, **kwargs):
             Additional keys and values to be written into dataset attrs.
         """
         self._parent = parent
+        super_kwargs = {k: kwargs[k] for k in ("compression", "shuffle") if k in kwargs}
```
ksunden, line 37:
```suggestion
```

initially thought passing kwargs here was needed, turned out only in the _data file

ksunden:
Dataset is (51, 61, 551) with array detector for the last axis, all other datasets are just 51x61

original file (without any compression): 34 MB
repacked: 29 MB
repacked and compressed: 1.8 MB

compression while collecting: 7.0 MB
repacked from compression while collecting (still compressed): 2.0 MB

kameyer226 (changes requested):
Win10 checks out

The test script needs ability to verify data integrity between uncompressed and compressed files.

untzag (approved):

ksunden:
honestly I'm not interested in testing data integrity, that is h5py's job, which I trust them greatly to do... The test is simply proving that it is indeed applied to begin with.

kameyer226:
ok

kameyer226 (approved):

PR #1088, opened 2022-07-25
Quick2d cmaps, and rotations of LinearSegmentColormaps via rotate_colormap

kameyer226:
## Changes

A method `rotate_colormap` was added to `wt.artists._colors`.   It takes a LinearSegmentedColormap and rotates it an integer number of times, returning a new LinearSegmentColormap.     It is not used for ListedColormaps.  Give it a try!

I allowed quick2D to use the `cmap` kwarg as it seems to work.

## Checklist

- [ ] added tests, if applicable  N/A
- [ ] updated documentation, if applicable N/A
- [X] updated CHANGELOG.md
- [ ] tests pass

### DIFF
--- a/WrightTools/artists/_colors.py
+++ b/WrightTools/artists/_colors.py
@@ -25,6 +25,7 @@ __all__ = [
     "grayify_cmap",
     "overline_colors",
     "plot_colormap_components",
+    "rotate_colormap",
 ]
 
 
@@ -139,7 +140,7 @@ def rotate_colormap(cmap, rotations, eps=0.0001):
     a new colormap.   Note:  these rotations are limited by the use of epsilon
     and may cause errors if the number of rotations is very large.  The last index value
     will be off by eps x rotations.   The method only works on tuples and arrays for segmentdata;
-    use the r parameter instead for make_cubehelix.
+    use the r parameter instead for make_cubehelix.   Not used for ListedColormaps.
 
     Parameters
     ----------
@@ -195,13 +196,13 @@ def rotate_colormap(cmap, rotations, eps=0.0001):
             for i in range(lnal):
                 arr = list(alphas[i])
                 arral.append(arr)
-            alphas = np.array(arral, dtpye=float)
+            alphas = np.array(arral, dtype=float)
     else:
-        reds = colors["red"]
-        greens = colors["green"]
-        blues = colors["blue"]
+        reds = np.array(colors["red"], dtype=float)
+        greens = np.array(colors["green"], dtype=float)
+        blues = np.array(colors["blue"], dtype=float)
         try:
-            alphas = colors["alpha"]
+            alphas = np.array(colors["alpha"], dtype=float)
             lnal = 1
         except:
             alphas = None
--- a/WrightTools/artists/_colors.py
+++ b/WrightTools/artists/_colors.py
@@ -133,6 +133,158 @@ def make_colormap(seq, name="CustomMap", plot=False):
     return cmap
 
 
+def rotate_colormap(cmap, rotations, eps=0.0001):
+    """
+    From a LinearSegmentedColormap, rotate an integer number of times, and return
+    a new colormap.   Note:  these rotations are limited by the use of epsilon
+    and may cause errors if the number of rotations is very large.  The last index value
+    will be off by eps x rotations.   The method only works on tuples and arrays for segmentdata;
+    use the r parameter instead for make_cubehelix.
+
+    Parameters
+    ----------
+    cmap : matplotlib.colors.LinearSegmentedColormap
+        Colormap to be cycled.
+    rotations : int
+        number of rotations
+    eps:  difference between end of first cycle and beginning of next cycle
+        which may need to be specified if the colormap has many index values
+
+    Returns
+    -------
+    matplotlib.colors.LinearSegmentedColormap
+
+    """
+    colors = cmap._segmentdata
+    newmap = cmap.copy()
+
+    if isinstance(colors["red"], tuple):
+        re = colors["red"]
+        gr = colors["green"]
+        bl = colors["blue"]
+        try:
+            alphas = colors["alpha"]
+            lnal = len(alphas)
+        except:
+            lnal = 0
+            alphas = None
+
+        lnre = len(re)
+        lngr = len(gr)
+        lnbl = len(bl)
+
+        arre = list()
+        argr = list()
+        arbl = list()
+
+        for i in range(lnre):
+            arr = list(re[i])
+            arre.append(arr)
+        for i in range(lngr):
+            arr = list(gr[i])
+            argr.append(arr)
+        for i in range(lnbl):
+            arr = list(bl[i])
+            arbl.append(arr)
+
+        reds = np.array(arre, dtype=float)
+        greens = np.array(argr, dtype=float)
+        blues = np.array(arbl, dtype=float)
+        if lnal != 0:
+            arral = list()
+            for i in range(lnal):
+                arr = list(alphas[i])
+                arral.append(arr)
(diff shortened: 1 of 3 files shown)


WrightTools/artists/_colors.py
```diff
@@ -133,6 +134,158 @@
...
+    newmap = cmap.copy()
+
+    if isinstance(colors["red"], tuple):
+        re = colors["red"]
+        gr = colors["green"]
+        bl = colors["blue"]
+        try:
+            alphas = colors["alpha"]
+            lnal = len(alphas)
+        except:
```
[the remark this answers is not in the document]
ddkohler: I think you are testing for a KeyError?  Don't allow any error to be excepted.

It's preferred to avoid try/except clauses if you can.  Here you can use if/else:  just look at the keys
`if "alpha" in colors.keys()`

WrightTools/artists/_colors.py
```diff
@@ -133,6 +134,158 @@ def make_colormap(seq, name="CustomMap", plot=False):
...
+
+    Returns
+    -------
+    matplotlib.colors.LinearSegmentedColormap
+
+    """
+    colors = cmap._segmentdata
+    newmap = cmap.copy()
+
+    if isinstance(colors["red"], tuple):
```
ddkohler, line 162: Do I understand correctly that you are just changing the tuple into numpy array?  e.g. can you not just use
`reds = np.array([ri for ri in colors["red"])`?

WrightTools/artists/_colors.py
```diff
@@ -133,6 +134,158 @@ def make_colormap(seq, name="CustomMap", plot=False):
...
+    """
+    From a LinearSegmentedColormap, rotate an integer number of times, and return
+    a new colormap.   Note:  these rotations are limited by the use of epsilon
+    and may cause errors if the number of rotations is very large.  The last index value
+    will be off by eps x rotations.   The method only works on tuples and arrays for segmentdata;
+    use the r parameter instead for make_cubehelix.   Not used for ListedColormaps.
+
+    Parameters
+    ----------
+    cmap : matplotlib.colors.LinearSegmentedColormap
```
ddkohler, line 147: I think you should make this work for any colormap, not just this specific type.

WrightTools/artists/_colors.py
```diff
@@ -133,6 +134,158 @@ def make_colormap(seq, name="CustomMap", plot=False):
     return cmap
 
 
+def rotate_colormap(cmap, rotations, eps=0.0001):
```
ddkohler, line 137: Consider more accurate names, e.g. `repeat_colormap`, `tile_colormap`

WrightTools/artists/_colors.py
```diff
@@ -133,6 +134,158 @@ def make_colormap(seq, name="CustomMap", plot=False):
...
+    will be off by eps x rotations.   The method only works on tuples and arrays for segmentdata;
+    use the r parameter instead for make_cubehelix.   Not used for ListedColormaps.
+
+    Parameters
+    ----------
+    cmap : matplotlib.colors.LinearSegmentedColormap
+        Colormap to be cycled.
+    rotations : int
+        number of rotations
+    eps:  difference between end of first cycle and beginning of next cycle
```
ddkohler, line 151: Can you say more about why this is needed?

WrightTools/artists/_colors.py
```diff
@@ -133,6 +134,158 @@ def make_colormap(seq, name="CustomMap", plot=False):
...
+    a new colormap.   Note:  these rotations are limited by the use of epsilon
+    and may cause errors if the number of rotations is very large.  The last index value
+    will be off by eps x rotations.   The method only works on tuples and arrays for segmentdata;
+    use the r parameter instead for make_cubehelix.   Not used for ListedColormaps.
+
+    Parameters
+    ----------
+    cmap : matplotlib.colors.LinearSegmentedColormap
+        Colormap to be cycled.
+    rotations : int
```
ddkohler, line 149: consider `repeats` or `cycles`, or `n`

WrightTools/artists/_colors.py
```diff
@@ -133,6 +134,158 @@ def make_colormap(seq, name="CustomMap", plot=False):
...
+        alphaf = alphas[-1, :][None, :]
+        alphavecr = alphavec / rotations
+        alphaveca = alphavecr
+
+    for i in range(rotations):
+        if i == 0:
+            red = np.append(redveca, redmat, axis=1)
+            green = np.append(greenveca, greenmat, axis=1)
+            blue = np.append(blueveca, bluemat, axis=1)
+            newmap._segmentdata["red"] = red
```
ddkohler, line 243: you are modifying hidden attributes (`_name`) which is not a supported aspect of matplotlib colormaps--they are subject to change without notice.  You should rely on the public properties of the colormaps to generate this repetition.  

I think this is also the reason your method is specific to the segmented type of colormaps?

ddkohler (changes requested):
Cool idea, @kameyer226 !  I think this code will be much more robust if you use the general framework of colormaps to implement this repeating aspect:

https://matplotlib.org/stable/api/_as_gen/matplotlib.colors.Colormap.html?highlight=colormap#matplotlib.colors.Colormap

e.g. a colormap can just be called (`cmap(np.linspace(0,1, 256)`) to generate colors as a well-defined type. You can just stack this array.

kameyer226:
I was able to mostly replicate the rotations method via the following code:

`import matplotlib.colors as clrs
clist=get_color_cycle(16,cmap="wright", rotations=3)
cm2=clrs.LinearSegmentedColormap.from_list(name="wrightcycle", clist)
wt.artists.quick2D(data, cmap=cm2)
`

This is where "rotations" comes from as a definition.   Is there a reason cmap is removed from the kwargs of quick2D?

WrightTools/artists/_quick.py
```diff
@@ -228,13 +229,17 @@ def quick2D(
...
     # colormap
     # get colormap
-    if data.channels[channel_index].signed:
-        cmap = "signed"
-    else:
-        cmap = "default"
-    cmap = colormaps[cmap]
-    cmap.set_bad([0.75] * 3, 1.0)
-    cmap.set_under([0.75] * 3, 1.0)
+    if isinstance(cmap, type(None)):
```
ddkohler, line 232:
```suggestion
    if cmap is None:
```

ddkohler:
your proposed solution:
```
import matplotlib.colors as clrs
clist=get_color_cycle(16,cmap="wright", rotations=3)
cm2=clrs.LinearSegmentedColormap.from_list(name="wrightcycle", clist)
wt.artists.quick2D(data, cmap=cm2)
```
is even better proposal than what I was thinking of. So are you saying that the solution is to just use that instead of introducing `rotate_colormap`? If not, we might not be on the same page...

> Is there a reason cmap is removed from the kwargs of quick2D?

I believe it was never "removed", it was just never implemented.  There was never a strong need to switch colormaps in quick2D.  Note that all our supported matplotlib methods (pcolor, contour, etc.) do support the cmap argument, it was never worked downstream.  I think it's a welcome change.

kameyer226:
I did not see the solution above until looking at the `from_list` method in more detail.   The proposed solution is to do this and close this PR, reopening another one just for the purpose of allowing cmap kwarg for quick2D.

On the other hand, the `rotations` does not appear to yield a full set of n rotations.   It is more like n-0.5 .   I do not know why it is doing this.  `rotate_colormap` does yield the full set of n rotations, however.

ddkohler:
> ...the `rotations` does not appear to yield a full set of n rotations. It is more like n-0.5 .

`n` refers to the total number of color elements that make up the cycled list--so if you want an integer number of cycles, make sure `n % rotations = 0`

kameyer226:
Closing

PR #1090, opened 2022-07-28
quick2d cmap added

kameyer226:
## Changes

Added `cmap` kwarg to the quick2D method. 

## Checklist

- [ ] added tests, if applicable  N/A
- [ ] updated documentation, if applicable  N/A
- [ X] updated CHANGELOG.md
- [ ] tests pass  N/A

### DIFF
--- a/WrightTools/artists/_quick.py
+++ b/WrightTools/artists/_quick.py
@@ -192,7 +192,7 @@ def quick2D(
         Dictionary of parameters in non-plotted dimension(s). If not
         provided, plots will be made at each coordinate.
     cmap : Colormap
-        Colormap to use.  If None, will use "default".
+        Colormap to use.  If None, will use "default" or "signed" depending on channel values.
     channel : string or integer (optional)
         Name or index of channel to plot. Default is 0.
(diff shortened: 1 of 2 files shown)


WrightTools/artists/_quick.py
```diff
@@ -228,13 +231,14 @@ def quick2D(
...
-    cmap.set_bad([0.75] * 3, 1.0)
-    cmap.set_under([0.75] * 3, 1.0)
+    if cmap is None:
+        if data.channels[channel_index].signed:
+            cmap = "signed"
+        else:
+            cmap = "default"
+        cmap = colormaps[cmap]
+        cmap.set_bad([0.75] * 3, 1.0)
+        cmap.set_under([0.75] * 3, 1.0)
```
ddkohler, line 241: It's probably best practice to consolidate this code block with `_parse_cmap` in the artists base module.  But the decisions in this code block are actually different than that parser, so some decisions need to be made before we do it.  

In any case, this change is not needed for this PR.  Just saying to keep it in mind.

ddkohler (approved):

PR #1092, opened 2022-07-29
chop respects axis order of args

### LINKED ISSUE #1091: chop: respect axes order in method call
The recent overhaul of chop left behind an unintended bug where the axis order in the call was not respected e.g.
```
data.chop("y", "x")[0]
print(data.axis_names)
```
This should return ("y","x"), but it might return ("x", "y"), depending on the original transform.

ddkohler:
## Changes

* resolves #1091 .

I am intentionally omitting a changelog entry since this bug is not present in any version.

## Checklist

- [ ] tests pass

### DIFF
--- a/WrightTools/data/_data.py
+++ b/WrightTools/data/_data.py
@@ -413,13 +413,14 @@ class Data(Group):
         """
         # parse args
         args = list(args)
-        transform_args = args.copy()
         for i, arg in enumerate(args):
             if isinstance(arg, int):
                 args[i] = self._axes[arg].natural_name
             elif isinstance(arg, str):
                 arg = arg.strip()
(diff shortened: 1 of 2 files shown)


ksunden (approved):

ksunden:
Well, I thought it looked good, but tests are failing...

Initially wondered if it would handle the case of being passed ints in, but we get everything as strings.

ksunden:
I think you took the opposite meaning from my comment...

I meant that I thought it was wrong initially, but upon closer inspection decided that it was correct _because_ we translate the ints to strings.

I did not think that was the error causing tests to fail...

ksunden (approved):

PR #1097, opened 2022-08-25
Update _axis.py

ddkohler:
remove axis whitespace on creation

## Changes

## Checklist

- [ ] updated CHANGELOG.md
- [ ] tests pass

### DIFF
--- a/WrightTools/data/_axis.py
+++ b/WrightTools/data/_axis.py
@@ -50,7 +50,11 @@ class Axis(object):
             Axis units. Default is None.
         """
         self.parent = parent
-        self.expression = expression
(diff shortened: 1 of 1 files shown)


ksunden:
Do we even care to warn here? I'm actually leaning no.

ddkohler:
My thought was that removing spaces could confuse a user if they define an expression with spaces and then try to find that expression later e.g.
```
expression = "a + b"
d.transform(expression)
expression in d.axis_expresssions # false
```

ddkohler:
If we don't think this will affect users, then I am also okay removing it.  Adding minor doc revisions to reflect the new expression behavior.

kameyer226 (approved):

Issue #1110, opened 2023-01-04
wt.data.join does not work if any pieces are a single point


DLafayetteII:
[test.txt](https://github.com/wright-group/WrightTools/files/10347261/test.txt)

I was confident I had done this successfully ~ a year ago, but this fails in WrightTools 3.4.4 and 3.4.6. 
Guessing something the slice inside the transpose:

`
  File "C:\Users\David Lafayette\miniconda3\lib\site-packages\WrightTools\data\_join.py", line 262, in join
    combine(data, out, channel_name, new_idx, transpose, slice_)

  File "C:\Users\David Lafayette\miniconda3\lib\site-packages\WrightTools\data\_join.py", line 199, in combine
    vals[valid_index] = old[:].transpose(transpose)[slice_]

ValueError: axes don't match array
`

DLafayetteII:
The issue arises from lines 245 - 250 of _join.py. np.ndim() of the 1 point variable returns a dimensionality of zero (np.ndim(d1.freq.points) = 0). Therefore, nothing is appended to the transpose, and an empty transpose (transpose = []) is passed to the combine method (If instead transpose = [0], the combine and join methods work as intended).

DLafayetteII:
I'll propose a fix for the issue but I'm not sure I understand the initial intention of what should be done with 'scalar' variables

PR #1111, opened 2023-01-04
1110 join single point error

DLafayetteII:
## Changes
Allowing join to succeed with data of shape (1,). Previously failed due to specific handling of data where np.ndim(data.variable) = 0. See issue #1110 for more info and the example failure case. Unsure of the motivation behind this specific handling of (1,) shape data, but this fix makes the join method work as expected.

## Checklist

- [ ] updated CHANGELOG.md
- [ ] tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 ## [Unreleased]
 
 ## Changed
+- Fixed join method to work with data of shape (1,)
 - `Axis`: space character ("\s") in expressions are culled.
 - fixed `interact2D` bug: channel/axes can now be specified with non-zero index arguments
 
--- a/WrightTools/data/_join.py
+++ b/WrightTools/data/_join.py
@@ -247,6 +247,7 @@ def join(
                 slice_.append(slice(None))
             # If p is scalar, a new axis must be added, no transpose needed
             else:
+                transpose.append(np.argmax(data[variable_name].shape))
                 slice_.append(np.newaxis)
             # Triple subscripting needed because newaxis only applys to numpy array
             # New axis added so that subtracting p will broadcast


kameyer226:
Code seems good, only move `transpose.append` statement in front of `if...else` since it is needed for both (all) cases, and remove the "no transpose needed" phrase

DLafayetteII:
> Just a moment, a new issue may appear with this fix. What if the variable of shape `(1,)` has a different value for each of the arrays being joined? If they are different, the join should fail, shouldn't it?

I am realizing that I have thus far only considered when the variable of shape (1,) is the transformed axis along which you are joining. In this case, it is fine if the values are different. If the shape (1,) variable is not your transformed axis along which you join, I want to say there's still no problem, but I have not tested and do not have a strong intuition.

kameyer226:
It looks good.

ksunden:
I'm a little confused as to how you got a scalar in the first place... I don't think we officially have supported such data objects, as their behavior is just different enough in all of these situations...

(Certainly you can have a size 1 data object with shape `[1,1,1]` for however many ndim you have, just thought scalar was not something we officially supported...)

ksunden:
note that scalar and shape `(1,)` are two different ideas, the latter will not have `np.ndim(var) == 0`

ksunden:
I would tend to want a test to show the failing behavior (and prevent it from recurring in the future... if we missed an edge case once, we could do so again)

kameyer226:
The variable was originally shaped `(1,)` prior to `join`, so it was never a scalar.    I tried joining data objects consisting of scalar variables and the attempt failed.

DLafayetteII:
> note that scalar and shape `(1,)` are two different ideas, the latter will not have `np.ndim(var) == 0`

This misconception is the heart of the issue. We do np.ndim(var.points), and np.ndim(var.points)==0, so scalar and shape(1,) get treated the same in the code.

PR #1128, opened 2023-06-21
guess_signed

### LINKED ISSUE #1124: kit function for guessing array plotting params
To be included in our workflow for plotting completed acquisitions, but as a kit function, it is available for e.g. "from" methods and other niche cases.  Analysis should determine guesses for signed/unsigned boolean, perhaps null, limits.

Heuristic based, using information like min/max, 95% window, or similar.

rpm4:
## Changes

* Added "signed" function for bluesky-cmds to create signed data. 
* resolves #1124 

## Checklist

- [x] updated CHANGELOG.md
- [x] tests pass


WrightTools/kit/_array.py
```diff
@@ -433,3 +434,21 @@ def enforce_mask_shape(mask, shape):
     """
     red = tuple([i for i in range(len(shape)) if shape[i] == 1])
     return mask.max(axis=red, keepdims=True)
+
+
+def signed(d0):
+    """Tells bluesky whether to sign data. We have a 7.5 % tolerance right now.%"""
```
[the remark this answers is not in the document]
ddkohler: The real comparison we want to make for this decision is to compare min and max to noise levels.  It would be nice to extract an estimate of noise for this evaluation.  But this is probably good enough for now.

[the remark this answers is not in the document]
rpm4: Just added some parameters, hopefully they look good

As to the second point: agreed! I think when we have more data it will be easier to obtain a better bound

WrightTools/kit/_array.py
```diff
@@ -433,3 +434,32 @@ def enforce_mask_shape(mask, shape):
     """
     red = tuple([i for i in range(len(shape)) if shape[i] == 1])
     return mask.max(axis=red, keepdims=True)
+
+
+def signed(d0):
+    """Tells bluesky whether to sign data. We have a 7.5 % tolerance right now.%
```
ddkohler, line 440:
```suggestion
    """Tells bluesky whether to sign data. We have a 7.5 % tolerance right now.
```

WrightTools/kit/_array.py
```diff
@@ -433,3 +434,32 @@ def enforce_mask_shape(mask, shape):
...
+    tolerance = (maxd0 - np.absolute(mind0)) / (maxd0 + np.absolute(mind0))
+
+    if tolerance < 7.5 * 10 ** (-2):
+        if min(d0) > 0:
+            return False
+        else:
+            return True
+
+    else:
+        return False
```
ddkohler, line 465:
```suggestion
    if tolerance >= 7.5e-2:
        return False
    else:
        return mind0 <= 0
```
More compact, perhaps more readable.  
The evaluation statements you have are actually booleans themselves (e.g. 1>=0 evaluates to True), so you don't have to turn them into True or False values--you can just return them.

tests/kit/signed.py
```diff
@@ -0,0 +1,22 @@
...
+
+import WrightTools as wt
+
+
+# --- test ---------------------------------------------------------------------------------------
+
+
+def test_5():
+    arr = np.array([-1, 0, 1])
+    assert wt.kit.signed(arr) == True
```
ddkohler, line 17:
```suggestion
    assert wt.kit.signed(arr)
```

tests/kit/signed.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def test_5_multiple():
+    arr = np.array([1, 3, 4, 11, 12])
+    assert wt.kit.signed(arr) == False
```
ddkohler, line 22:
```suggestion
    assert not wt.kit.signed(arr)
```

WrightTools/kit/_array.py
```diff
@@ -202,7 +203,7 @@ def orthogonal(*args) -> bool:
         if hasattr(arg, "shape"):
             args[i] = arg.shape
     for s in zip(*args):
-        if np.prod(s) != max(s):
+        if np.product(s) != max(s):
```
ddkohler, line 206:
```suggestion
        if np.prod(s) != max(s):
```

PR #1134, opened 2023-07-07
`Data.squeeze`

### LINKED ISSUE #1069: Data.chop: add "drop" kwarg to bypass MultidimensionalAxisError
Looking for easy method for chopping to remove multidimensional array dims we do not care about.
This is based on the issue of accidentally recording superfluous data on an array detector, when all we care about is the non-array data and axes.

e.g. we have interesting channels/variables of shape (n,m,1), but uninteresting channels/variables of shape (n,m,p) that we wish to dispose of.  

@ksunden had a good proposal:
> perhaps we could add a "drop" kwarg to chop which would make this
>```
>data.transform("w1_points", "d1_points")
>data = data.chop("w1_points", "d1_points", drop=("wavelengths", "intensities"))
> ```
> possibly could even be
> ```
> data.transform("w1_points", "d1_points")
> data = data.chop("w1_points", "d1_points", drop=True)
> ```
> where drop=True means "drop all vars and channels which have non-one in the shape where the chopped axes have one in their shape" (perhaps could be worded more succinctly, but that is the idea)

ddkohler:
## Changes

* new Data method `squeeze`, our analogue of np.squeeze
* resolves #1069 

## Checklist

- [x] updated CHANGELOG.md
- [x] tests pass


tests/data/squeeze.py
```diff
@@ -0,0 +1,32 @@
+#! /usr/bin/env python3
+"""Test chop."""
```
kameyer226, line 2: -> """Test squeeze."""

kameyer226 (changes requested):
Should be good except for the one replace.  Nice addition!

ddkohler:
This PR is still a work in progress.  Specifically, the squeezed data objects need to work on the Constant objects they create.

kameyer226 (approved):

PR #1139, opened 2023-08-10
Docs

ddkohler:
## Changes

avoid deprecation of Read the Docs

[builds on readthedocs are currently failing](https://readthedocs.org/projects/wrighttools/builds/23210803/), so this is an attempt to remedy this


pcruzparri (commented):
Reading the readthedocs documentation, I think that what we had would still work. However, I provided an alternative in case the docs/requirements.txt is necessary. Additionally, from one of the examples on their documentation of the requirements file, another way is to add an install option to python as shown below: 

```
python:
  version: 3.8
  pip_install: true
  install:
    path: ..
    extra_requirements: 
      - docs
```
which should similarly run `pip install ..[docs]` during building and is what is needed. 

Thoughts?

ddkohler:
I will try to mirror the attune configs setup, since they are still working well.

kameyer226 (approved):

PR #1150, opened 2024-01-26
safeguard join "method" kwarg

### LINKED ISSUE #1118: wt.data.join throws no error when method argument is not an expected string
Method argument only does something with methods {"first", "last", "min", "max",  "mean"}. If you provide a different string, i.e. "average", "fkjgkjfd", etc.,  wt.data.join throws no error and outputs something. 

I have not checked what results from a data.join without a proper method, but we should throw an error to avoid confusion, like in the case where a user uses "average" instead of "mean".

### LINKED ISSUE #1119: wt.data.join claims method = 'sum' is a valid choice in docstring, but method = 'sum' not implemented
If there is a potential use case for 'sum', we should implement. Else we should remove from documentation.

ddkohler:
## Changes

* allowed methods are explicitly checked:  min, max, mean, first, last
* resolves #1118
* resolves #1119

## Checklist

- [ ] updated CHANGELOG.md
- [ ] tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 ### Fixed
 - `interact2D`: fixed bug where use_imshow broke the sliders
 - `data.join` ensures valid `method` is selected
+
+### Changed
 - `data.join` no longer supports `sum` method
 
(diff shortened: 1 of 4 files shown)


kameyer226 (approved):
Sorry that sum didn't work out...

PR #1156, opened 2024-02-22
Fix build

ddkohler:
For the near term, this is my final attempt to fix the pdf builds.  I found I had some bad rendering still from improper rst writing (my bad from the last pr).

Since html hasn't had any problems building (just the pdf), I propose we turn off pdf renders if this issue persists.  html is rendering just fine.  If I don't learn anything with this build, I will turn off pdf builds.

It's hard to run our full sphinx workflow on windows, especially with the latex.  I made some changes to the docs builds that make it a little more friendly for windows.

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -5,13 +5,13 @@ Command Line Interface (CLI)
 
 Most of the CLI interface is new as of v3.5.2 and under active depelopment.
 The commands here may change.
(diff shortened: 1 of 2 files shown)


kameyer226 (approved):

ddkohler:
Okay, it looks like there are still (several) build issues with the pdf.  In the near term, I will remove pdf rendering, and make an issue.  This will allow html docs to build.

Issue #943, opened 2020-05-24
consider utilizing gif


untzag:
We could get rid of our builtin / hacky animated gif support.

https://github.com/maxhumber/gif

darienmorrow:
Here is an example usage of this library for a simulation of @ddkohler.   

![wigners](https://user-images.githubusercontent.com/15792657/82763980-1092c000-9dd1-11ea-9ff4-3ed26973e6b5.gif)

```
import gif
import matplotlib.pyplot as plt
import WrightTools as wt
import numpy as np

chan = "dI_I"
p = "tdle_sim.wt5"
d = wt.open(p)
d.create_channel(name="dI_I", values=d.A[:] + d.B[:] + d.C[:] + d.D[:], signed=True)
for channel in d.channels:
    channel.normalize()
    
@gif.frame
def plot_wigner(pump):
    fig, gs = wt.artists.create_figure(cols=[1,'cbar'])
    ax = plt.subplot(gs[0])
    dx = d.chop("w_probe", "delay", at={"w_pump": [pump, "eV"]}, verbose=False)[0]
    dx.constants[0].format_spec = ".2f"
    label = dx.constants[0].label
    label = label.replace("w\;pump", "\\hslash\\omega_{pump}")  
    ax.contourf(dx, channel=chan, vmin=-1, vmax=1)
    ax.contour(dx, levels=np.array([0]), colors='k', channel=chan,)
    ax.set_title(label, fontsize=18)
    ax.grid()
    xlabel = '$\\mathsf{\\hslash\\omega_{probe} \\; (eV)}$'
    ylabel = '$\\mathsf{T \\; (ps)}$'
    wt.artists.set_fig_labels(xlabel=xlabel, ylabel=ylabel)
    cax = plt.subplot(gs[1])
    ticks = np.linspace(-1,1,11)
    label = '$\\mathsf{\\Delta I}$'
    wt.artists.plot_colorbar(cax=cax, cmap='signed', ticks=ticks, label=label)

    
frames = []
for pump in d.w_pump.points:
    frame = plot_wigner(pump)
    frames.append(frame)

gif.save(frames, "wigners.gif", duration=100)
```

[tdle_sim.zip](https://github.com/wright-group/WrightTools/files/4674325/tdle_sim.zip)

darienmorrow:
The way the color space is segmented is a bit problematic...

ksunden:
[...]

PR #1229, opened 2026-03-12
animation helpers

### LINKED ISSUE #1233: artists.quick:  iterator implementation
quick1D and quick2D have been steadfast tools for getting nicely formatted plots quickly from datasets.  When the plotters are given higher dimensional data, they will chop and create a figure for each item.  The problem I am bringing up here is that, historically, these methods have made the figures all in one go (i.e. as a Collection).  Iterators are less intense on resources (only one figure is made at a time).  In general, this seems like a better option, especially with cameras are becoming more prominent channels in out multidimensional datasets.  Iterators also naturally integrate with matplotlib's `animation` module.

I propose new functions `quick2Ds` and `quick1Ds` to implement this, e.g. something like:

```python
def quick2Ds(data, ...):
    fig = initialize_figure()
    for di in data.ichop(...):
         yield update_figure(fig, di)
```

ddkohler:
## Changes

The intent here is to wrap `matplotlib.animations.FuncAnimate` so that it can handle data objects, interact2D figures, or quick[1/2]D-formatted figures and produce animations.  

To make this work with quick plotters, a significant refactor was done. 

* addresses #943
* resolves #1233 

### Added
* `animate.animate2D`: versatile conversion of data to animation; lots of matplotlib customizations
* `animate.animate_interact2D`:  create an animation from an interact2D object.  The animation walks all the sliders.
* `animate_quick`: create an animation whose frames are the figures that would be created in a quick1D or quick2D call
* `Quick2DIterator`, `Quick1DIterator`: iterator classes for making figures for each object in a data chop.
* arists: `quick2Ds`, `quick1Ds`:  functions that generate the quick iterator classes using explicit arguments.

### Changed
* `interact2D` has a static output type (rather than a SimpleNamespace instance) to clearly convey the inputs for animate_interact2D.
* `quick1D`, `quick2D`:  refactored for integration with iterators, animations (these functions are wrappers for the class `Quick1DLegacy`, `Quick2DLegacy`).

### TODO
* [x] ~`wt5 animate` entry point~ discarding; can be added at later time
* [x] ~incorporate snaking (for 4+ dimensions)~ discarding as a low priority
* [x] incorporate back-and-forth option (to avoid jagged animation endings)
* [x] ~`animate2D`: options for plotting something other than the final two axes~ discarding; can be added at later time
* [x] refactor `artists.quick` (minimize redundancy between iterator- and list-based functions)

## Checklist

- [x] updated CHANGELOG.md
- [x] tests pass


ddkohler:
It looks like this is far enough along that we can merge now and deal with issues as they come.

PR #1252, opened 2026-09-02
update from_solis

### LINKED ISSUE #1249: from_Solis: acceptance of both top and bottom appended attrs in .asc file format
When exporting from Andor Solis native .sif format to the .asc format that WrightTools expects, comments within the from_Solis method advise users to opt-in to appending the acquisition attributes to the file. When you elect to do so, Andor Solis asks for whether they should be appended to the top or bottom of the file. Our current WrightTools.data.from_Solis expects data at the top, and checks for attrs at the bottom. The method fails if the attrs are instead appended at the top of the file. 

Given that the top vs bottom choice is arbitrary, and top is at least the default in the current version of Andor Solis I am using, it would be nice to allow for both export formats. We could make this an argument of the method (default can be bottom to maintain previous methods), or since the formatting of the attrs is different from the data, we could build in a check to automatically determine whether data or attrs is listed first, treating accordingly.

ddkohler:
Allow metadata to be at the top or bottom of the file.

## Changes

- resolves #1249 

## Checklist

- [x] updated CHANGELOG.md
- [x] tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Fixed
+- `from_solis`:  accepts metadata listed before or after data
+
 ## [3.6.4]
 
 ### Added
--- a/WrightTools/data/_solis.py
+++ b/WrightTools/data/_solis.py
@@ -7,7 +7,7 @@ import os
 import pathlib
 import time
 import warnings
-
+import string
 import numpy as np
 
 from ._data import Data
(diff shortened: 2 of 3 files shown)


DLafayetteII:
Tested on one file with attrs appended by Andor Solis export on either top or bottom. Worked in both cases, and attrs were populated and equivalent in both cases (besides name/source as those are dependent on file naming). 

Note that this workup would still fail using this merged branch for my specific data, due to Pull request #1247 still being outstanding. I think that one is outstanding due to me not filling out the changelog, so will go investigate.

